PR #19, opened 2016-02-11
This might be the new master?

ionefine:


python/electrode2currentmap.py
```diff
@@ -95,15 +98,99 @@ def current_spread(self, xg, yg, alpha=14000, n=1.69):
         return np.sum(c, 0)
 
 
-class Stimulus(TimeSeries):
+def receptive_field(electrode, xg, yg, size):
+        """
+        # TODO currently this is in units of the grid, needs to be converted to microns
```
arokem: I think this is in the correct units now?

python/electrode2currentmap.py
```diff
@@ -95,15 +98,99 @@ def current_spread(self, xg, yg, alpha=14000, n=1.69):
...
+        """
+        rf=np.zeros(xg.shape)
+        ind=np.where((xg>electrode.x-(size/2)) & (xg<electrode.x+(size/2))
+         & (yg>electrode.y-(size/2)) & (yg<electrode.y+(size/2)))
```
arokem: Could you indent this a bit? Would make it easier to read

python/electrode2currentmap.py
```diff
@@ -95,15 +98,99 @@ def current_spread(self, xg, yg, alpha=14000, n=1.69):
...
+    """
+    Is used to generate pulse trains to simulate psychophysical experiments.
+    
     """
 
     def __init__(self, freq=20, dur=0.5, pulse_dur=.075/1000.,interphase_dur=.075/1000., delay=0.,
                  tsample=.005/1000., current_amplitude=20,
                  current=None, pulsetype='cathodicfirst', stimtype='pulsetrain'):
         """
+        Ariel, do we need the variable current?
```
arokem: Nope. We can eliminate that, I think.

arokem:
Cool. I had a couple of comments, but I think we can do that in subsequent work.

PR #22, opened 2016-02-19
The Dorn experiment now happens pixel by pixel.

arokem:


arokem:
Still having issues with the output of oyster relative to the meshgrids. We
need to sort that out: if the meshgrids are initialized with 'xy' indexing,
they seem to have the wrong shapes, but if they're initialized with 'ij'
indexing, the oyster code doesn't do anything (the effective current map is
the the same as the "passive spread" current map). My sense is that we need
to use 'ij' indexing, but change the code in oyster to take that into
account. I figure the way forward is to write a test for oyster, and then
refactor it to use the meshgrids with 'ij' indexing.

ionefine:
I’m having a weird issue where it’s still not running at all, I accepted the pull,  downloaded the master, and made my code look like your Exp_Dorn11 notebook. Could there be a version issue where you are using a different version of effectivecurrent2brightness or electrode2current map?

Ione

arokem:
There shouldn't be -- unless you have some local changes in these files (do
you?). What error are you seeing?

ionefine:
Dojit!

Duh!

Ione

arokem:
Ah yes. I changed that one to True...

arokem:
BTW - I am currently running a full Dorn thing with all 60 electrodes and a
retina with 5000 by 5000 microns. Might take a while...

ionefine:
So am I! Maybe I’ll stop.

Do we still need to swop the ij/xy. G & I can work on that.

Ione

arokem:
Yeah - I don't entirely understand the ij/xy story fully yet.

I do know that when we swapped to 'ij', the oyster model gave outputs that
were quite different. Sadly, it also failed silently (as in, the ecm and cm
were identical).

I think that ij is what we want. I tried making oyster play along last
night, but couldn't get it to work.

arokem:
FWIW, my Dorn script is still running...

PR #23, opened 2016-02-23
Working to speed up movie generation

arokem:
Started with adding some testing. Will add more tests for the relevant bits of code before refactoring.


arokem:
This is ready to be merged, I believe. A lot of this is clean-up, and some testing implemented to pave the way towards a refactoring of the stimulus generation functions, that will hopefully speed things up a bit.

ionefine:
Will do it tomorrow when haven’t had a glass of wine unless you don’t think I need to look at it before merging.

Running some more movies with a larger lambda and the right amplitude scaling
Ione

arokem:
Great! I'll keep the AWS machine running.

PR #38, opened 2016-03-10
Various changes including a script for converting and image into a movie

ionefine:
Added a script called ConvertImage2Percept that turns an jpg into an array and from there to a movie and so on ...

python/utils.py
made it number of processors -1 in an attempt to not have it wedge things.

Electrode2currentmap This is the only possibly important change. There was a weird thing where the max current in the original current spread was 1 (under the electrode) but, because of summation over the electrode it was much higher (17) in the 'current spread that included axonal integration I’ve added a normalization step 
ecs=ecs /ecs.max() to deal with that, so the max of ecs is now also 1. That prevents sensitivity changing with lambda.

As part of that normalization step I’ve also lowered the max amplitude a touch (at one point I was really confused that I seemed to have to lower the amplitude to much lower than I was expecting to get the model to work – that was why). Anyway that amplitude parameter is something to think about. At the moment it is 2x threshold.

Less important changes are changing the default parameters so it uses a crude temporal resolution and a .45 pulse width as the default. The default should certainly be the 0.45 pulse width

effectivecurrenttobrightness
Instead of calculating rs each time, I have it calculate it in the script and then send in rs instead of fps

Calc_pixel: just swopped the order of the inputs in a matlab neurotic way. Purely aesthetic.


python/utils.py
```diff
@@ -133,6 +133,7 @@ def parfor(func, in_list, out_shape=None, n_jobs=-1, func_args=[],
     """
     if n_jobs == -1:
         n_jobs = multiprocessing.cpu_count()
+        n_jobs=n_jobs-1
```
arokem: Great. Might not help with memory issues, but at least it will (hopefully) leave one CPU available to log into the machine and shut things down.

python/effectivecurrent2brightness.py
```diff
@@ -145,7 +145,7 @@ def pulse2percept(temporal_model, ecs, retina, stimuli,
     subsample_factor : float/int, optional
     dojit : bool, optional
     """
-    rs = int(1 / (fps*stimuli[0].tsample))
```
arokem: I am not sure this change is necessary. It only gets calculate once in here.

arokem:
I had that one comment, but I'll go ahead and merge this, and we can rethink later, as needed.

PR #46, opened 2016-09-13
Profiling pulse2percept

mbeyeler:
This pull request makes the model run ~45% faster and adds support for the "Krishnan" flavor of the model.

New features:
* Two flavors of the model: `'Nanduri'` and `'Krishnan'`, based on Ione's code. The two flavors are mathematically equivalent. However, in practice there might be small numerical differences.
   * `tm_nanduri = effectivecurrent2brightness.TemporalModel(model='Nanduri')`: first calculates the fast response, followed by the charge accumulation step.
   * `tm_krishnan = effectivecurrent2brightness.TemporalModel(model='Krishnan')`:  first performs the charge accumulation step, followed by the fast response.
* New multiprocessing engine `dask`, based on Ariel's `pyAFQ` code. The alternative (and default) is `joblib`.

Profiling:
* Localized and removed memory leak issues (TimeSeries constructor)
* Achieved speedup of ~45% for typical model parameters (20nA/20Hz pulse, 0.1-0.5 stimulus duration, 5e-6s integration step).
* `python/notebooks/0.0-mb-profiling-temporal-model.ipynb`: Illustrates the performance improvement for every stage of the temporal model cascade.
* `python/notebooks/0.1-mb-profiling-pulse2percept.ipynb`: Demonstrates the speedup and scalability of the whole pipeline, compared to the current master branch. The check is performed against the current code base, so can be used for future performance checks, too.


python/electrode2currentmap.py
```diff
@@ -279,17 +279,23 @@ def __init__(self, freq=20, dur=0.5, pulse_dur=.075/1000.,
 
         # set up the sequence
         if stimtype == 'pulsetrain':
-            interpulsegap = np.zeros(round((1/freq) / tsample) - len(pulse))
+            # Warning: using a non-integer number instead of an integer will
+            # result in an error in the future
+            interpulsegap = np.zeros(int(1/freq/tsample) - len(pulse))
```
arokem: `int` will always chop down to the smaller integer (e.g., `int(1.8) == 1`). Is that really what we want here?

PR #50, opened 2016-10-17
TST: Added a travis configuration file.

arokem:


arokem:
OK - it seems to install fine now, and tests pass on Python 3, but not on Python 2. There are also plenty of PEP8 violations that need to be fixed up.

PR #58, opened 2017-01-09
First pass at setting up documentation.

arokem:
Not everything is in place quite yet...


arokem:
TODO: 

- [ ] Place the videos properly on the front page. 
- [ ] Examples folder in top level with a couple of example scripts. 

Once everything is in place, we can run `make upload` to get the docs into the gh-pages branch

arokem:
Looks great: http://uwescience.github.io/pulse2percept

The one thing that's still a bit weird is that this page has nothing on it: http://uwescience.github.io/pulse2percept/reference/pulse2percept.html

It looks like it's not picking up the functions in that module :-/

mbeyeler:
Well, there is no such module, right? That is, we don't have a high-level `pulse2percept` module, only a `pulse2percept` function in the `effectivecurrent2brightness` module.

Now the Python 3 tests seem to get stuck somehow and then error out... Python 2 tests run fine in my repo. Not sure what's going with Python 3...
There's also a PEP complaint but that's an easy fix.

arokem:
Yeah - got it. But it's weird that there's this empty page then.

I guess if we put a top-level docstring here:
https://github.com/uwescience/pulse2percept/blob/master/pulse2percept/__init__.py#L1

it would get rendered in here...

arokem:
I think this is ready to go!

PR #59, opened 2017-01-11
REVIEW mb/fixElectrodeArray

mbeyeler:
A few improvements for the ElectrodeArray class and some helper functions. Most of these are already being used in my experiment branches in one form or the other, but are now also tested and integrated. So the purpose of their existence is mostly to make my life easier. 😸 

Argus I arrays now support legacy naming (L1, L2, ..., L8, M1, M2, ..., M8).

The electrode-to-retina distance depends on electrode type and electrode location. This computation has been moved to `Electrode.set_height` so that heights can be overridden.

Keeping track of electrode names in `ElectrodeArray.names` is redundant, because every `Electrode` has a `name` itself. Setting custom (or new) names would have to be done in both places, otherwise indexing might break. Therefore, I removed `ElectrodeArray.names`. In order to do the indexing, the loop has to be over `name` of every `Electrode` in an `ElectrodeArray`. In order to speed up the average lookup time, the array is traversed in random order.

The parameter `tol` of `pulse2percept` now specifies a fraction instead of an absolute value: For example, `tol=0.1` will discard all pixels whose effective current is <10% of the max effective current value.

I also added a helper function that can find files in a directory whose names match a regular expression pattern.

### DIFF
--- a/pulse2percept/files.py
+++ b/pulse2percept/files.py
@@ -58,3 +58,35 @@ def scale(inarray, newmin=0, newmax=1):
     delta = (newmax - newmin) / (oldmax - oldmin)
     outarray = delta * (inarray - oldmin) + newmin
     return outarray
+
+
+def find_files_like(datapath, pattern):
+    """Finds files in a folder whose name matches a pattern
+
+    This function looks for files in folder `datapath` that match a regular
+    expression `pattern`.
+
+    Parameters
+    ----------
+    datapath : str
+        Path to search
+    pattern : str
+        A valid regular expression pattern
+
+    Examples
+    --------
+    # Find all '.npz' files in parent dir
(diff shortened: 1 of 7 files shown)


pulse2percept/files.py
```diff
@@ -58,3 +58,35 @@ def scale(inarray, newmin=0, newmax=1):
...
+    >>> files = find_files_like('..', '.*\.npz$')
+    """
+    # No need to import these at module level
+    from os import listdir
+    import re
+
+    # Traverse file list and look for `pattern`
+    filenames = []
+    pattern = re.compile(pattern)
+    for file in listdir(datapath):
```
arokem: Please change `file` to be something that isn't a reserved name in Python (`fname` would work)

PR #61, opened 2017-01-31
mb/logging

mbeyeler:
Add a basic logger to replace in-code print statements.

Prints messages >= `INFO` to console and stores messages >= `DEBUG` in local file `debug.log`.


arokem:
:+1:

PR #68, opened 2017-02-21
[MRG] mb/refactor

mbeyeler:
Refactoring the code: The public user interface now lives in an `api` module that gets imported into the `pulse2percept` namespace. Relevant function calls are wrapped by a `Simulation` class.

Example usage:
```
import pulse2percept as p2p
import matplotlib.pyplot as plt

# Electrode arrays now live in the `implant` module
argus = p2p.implants.ArgusII()

# Stimuli now live in the `stimuli` module
pt = p2p.stimuli.biphasic_pulse(freq=50, amp=20, tsample=0.005/1000)
stim = {'C4': pt}

# Most work happens via the Simulation class
percept = sim.Simulation("example", argus, engine='joblib')

# Specify parameters via setter methods for the different layers:
# optic fiber layer (OFL), ganglion cell layer (GCL), and inner
# nuclear layer (INL).
# If not called, layers get initialized with default values upon
first call to sim.pulse2percept.
sim.set_ofl(lambda=2)

# Run the simulation
percept = sim.pulse2percept(stim, tol=0.05)

# Plot
frame = p2p.get_brightest_frame(percept)
plt.imshow(frame.data, cmap='gray')

# etc.
sim.plot_fundus(stim)
sim.save()
```

Modules:
- `api`: public user interface
- `retina`: has `Retina`, `TemporalModel`, `oyster`, etc.
- `implants`: has `Electrode`, `ElectrodeArray`, `ArgusI`, etc.
- `stimuli`: has `biphasic_pulse`, `Psycho2Pulsetrain`, `image2pulsetrain`, etc.
- `utils`: has `TimeSeries`, `gamma`, `sparseconv`, etc.
- `files`: has file I/O

Tasks:
- [X] Create `api` module
- [X] Create `retina` module
- [X] Add Simulation class
- [X] Add setters for model specs
- [X] Add deepcopy to `TimeSeries`
- [X] Move and improve `gamma`
- [X] Add tests for all new functionality
- [X] Document all new functionality


pulse2percept/api.py
```diff
@@ -0,0 +1,423 @@
...
+
+        # Perform any necessary calculations per electrode
+        pt_list = p2p.utils.parfor(self.gcl.calc_per_electrode,
+                                   pt_list, engine=self.engine,
+                                   n_jobs=self.num_jobs)
+
+        # Which layer to simulate is given by implant type.
+        # For now, both implant types process the same two layers. In the
+        # future, these layers might differ. Order doesn't matter.
+        if self.implant.etype == 'epiretinal':
```
arokem: Add an input for `layers` which will allow you to bypass these altogether. That way, you can input (for example): `layers = ['NFL']` (which you currently can't).  

Additionally, add a 'GCL' layer, for the ganglion cells. That way, our first set of simulations will pertain to a model that has `layers=['NFL', 'GCL']` (which is neither of the currently existing options). 

This also allows us to leave the current `epiretinal` and `subretinal` options and keep their functionality

pulse2percept/api.py
```diff
@@ -0,0 +1,423 @@
...
+        if t_percept != percept.tsample:
+            percept = percept.resample(t_percept)
+
+        logging.getLogger(__name__).info("Done.")
+
+        return percept
+
+    def set_ofl(self, sampling=100, axon_lambda=2, rot_deg=0,
+                x_range=None, y_range=None,
+                loadpath='../', save_data=True, streaks_enabled=True):
```
arokem: This `loadpath` is a bit odd. Consider something more rational (e.g., `~/.pulse2percept` or `.`)

pulse2percept/api.py
```diff
@@ -0,0 +1,423 @@
...
+        if self.ofl_streaks:
+            ecs, _ = self.ofl.electrode_ecs(self.implant)
+        else:
+            _, ecs = self.ofl.electrode_ecs(self.implant)
+
+        # Calculate the max of every current spread map
+        lmax = np.zeros((2, ecs.shape[-1]))
+        if 'INL' in dolayers:
+            lmax[0, :] = ecs[:, :, 0, :].max(axis=(0, 1))
+        if 'NFL' in dolayers:
```
arokem: Maybe we can start moving towards using "OFL" instead of "NFL"?

`if 'NFL' in dolayers or 'OFL' in dolayers:`

etc.

Preferably remove `NFL` altogether. We don't need to worry about back compatibility for this -- to internal for our current users.

pulse2percept/stimuli.py
```diff
@@ -0,0 +1,531 @@
...
+        ppt = ppt[0:int(round(dur / tsample))]
+        intfunc = spi.interp1d(np.linspace(0, len(rflum), len(rflum)),
+                               rflum)
+
+        amp = intfunc(np.linspace(0, len(rflum), len(ppt)))
+        data = amp * ppt * amp_max
+        utils.TimeSeries.__init__(self, tsample, data)
+
+
+class Psycho2Pulsetrain(utils.TimeSeries):
```
arokem: Rename this to `PulseTrain` and keep pretty much as is.

Rename `mono` and `biphasic` pulses to `Mono` and `BiphasicPulse` and make them classes, instead of functions. 

In other words : separate what should be classes and what should be functions (`movie2pulsetrain` and `image2pulstrain` can remain functions that generate `PulseTrain` objects).

PR #69, opened 2017-03-07
[MRG] Document files module

mbeyeler:
Add documentation and testing of the `files` module:

- [X] Document functions
- [X] Test functions
- [X] Trigger `ImportError`
- [X] Improve coverage

### DIFF
--- a/pulse2percept/files.py
+++ b/pulse2percept/files.py
@@ -5,7 +5,6 @@ import os
 import logging
 
 from pulse2percept import utils
-from pulse2percept import electrode2currentmap as e2cm
 
 
 @utils.deprecated
--- a/pulse2percept/files.py
+++ b/pulse2percept/files.py
@@ -5,13 +5,15 @@ import os
 import logging
 
 from pulse2percept import utils
+from pulse2percept import electrode2currentmap as e2cm
 
 
 @utils.deprecated
 def savemoviefiles(filestr, data, path='savedImages/'):
     """Saves a brightness movie to .npy, .mat, and .avi format
 
-    This function is deprecated as of v0.2.
+    This function is deprecated as of v0.2 and will be removed
+    completely in v0.3.
 
     Parameters
     ----------
@@ -24,6 +26,8 @@ def savemoviefiles(filestr, data, path='savedImages/'):
     path : str, optional
         Path to directory where files should be saved.
         Default: savedImages/
+
+    .. deprecated:: 0.2
     """
     np.save(path + filestr, data)  # save as npy
     sio.savemat(path + filestr + '.mat', dict(mov=data))  # save as matfile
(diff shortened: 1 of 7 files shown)


pulse2percept/utils.py
```diff
@@ -13,8 +16,27 @@
     has_jit = False
 
 
+def deprecated(func):
```
arokem: Maybe add a key-word argument for "alt_func", which, if defined, will trigger a block that also prints a `"Use {} instead".format(alt_func.__name__)`?

pulse2percept/tests/test_files.py
```diff
@@ -1,5 +1,57 @@
...
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from pulse2percept import files
+
+
+def test_savemoviefiles():
+    # This function is deprecated
```
arokem: Do you want to test that the warning is raised?

mbeyeler:
The nested decorator turned out to be a bit of a nightmare. [This advice](http://stackoverflow.com/questions/35572663/using-python-decorator-with-or-without-parentheses) turned out to be the best one (also [this](https://github.com/pytest-dev/pytest/blob/2e04771893f17b571e6ab48f6884a6a70cde8fc8/_pytest/python.py#L132)).

It works fine if you do `@deprecated(arg=p2p.utils.TimeSeries)` or something. Problem is if you do `@deprecated`, without passing `arg`, the current function object gets assigned to `arg`, still. So there's no way to differentiate whether `arg` is the current function object or the alternative function object.
Solution: Don't pass the object of the alternative function, use a string instead.

Ready to merge.

pulse2percept/files.py
```diff
@@ -1,31 +1,71 @@
...
+    Parameters
+    ----------
+    filename : str
+        File name of .avi movie to be produced
+    movie : array
+        A 3-D NumPy array containing the data, such as the `data` field of
+        a utils.TimeSeries object
+    rate : float, optional
+        Frame rate. Default: 30 Hz
+    """
```
arokem: Please add the `.. deprecated:: 0.2` tag here.

pulse2percept/utils.py
```diff
@@ -13,8 +14,55 @@
...
+        # Direct decoration
+        return wrap(arg)
+    else:
+        # Alternative function given
+        alt_func = arg
+        return wrap
+
+
 class Parameters(object):
+    """Container to wrap a MATLAB array in a Python dict"""
```
arokem: Add a note that this is going to be deprecated in 0.2

mbeyeler:
Great point about the `arg`. I removed it from kwargs and made it an arg instead. I also added the deprecated flags you mentioned.

pulse2percept/utils.py
```diff
@@ -42,7 +42,7 @@ def wrap(func):
         """
         e_s = "Call to deprecated function %s." % func.__name__
         if alt_func:
-            e_s += "Use %s instead." % alt_func
+            e_s += " Use %s instead." % alt_func
```
arokem: Sorry - I think I missed this before. Where does `alt_func` come from? 

It doesn't seem to be defined anywhere. I must be missing something

mbeyeler: It is set outside `wrap()` but inside `deprecated()` (starting l.57), in the case where `arg` is not callable:
```
alt_func = None
if callable(arg):
    # Direct decoration
    return wrap(arg)
else:
    # Alternative function given
    alt_func = arg
    return wrap
```

arokem:
I knew I was missing something.

PR #72, opened 2017-03-21
[MRG] mb/video2pulsetrain

mbeyeler:
This PR adds a `video2pulsetrain` function to convert video files into stimuli. Videos can be loaded from file, and percepts can be stored to file. Uses Scikit-Video with either an `ffmpeg` or `libav` backend. Replaces some deprecated functionality such as `p2p.files.savemoviefiles` and `p2p.stimuli.Movie2Pulsetrain`.

- [X] Add `video2pulsetrain`
  - [X] processes every frame with `image2pulsetrain`
  - [X] supports both amplitude and frequency modulation
  - [X] makes `Movie2Pulsetrain`, `receptive_field`, and `retinalmovie2electrodtimeseries` obsolete
- [X] Add load/save utilities to `files` module
  - [X] `load_video` loads a video file either into a NumPy array or a TimeSeries object
  - [X] `save_video` stores either a NumPy ndarray or a TimeSeries object in a video file
  - [X] `save_video_sidebyside` stores both an input video and the `TimeSeries` output in a video file
  - [X] makes `savemoviefiles` and `npy2movie` obsolete
- [X] Add easy concatenation of pulse trains `pt1.append(pt2)`
- [X] Switch Travis build to `trusty` distro
- [X] Fix `gap_size` issue in `image2pulsetrain`
- [X] Fix `effectivecurrent` calculation when not all layers selected
- [X] Add tests that process example images/videos from Scikit-Learn/Scikit-Video
- [X] Complete documentation of new functionality
- [X] Add updated `image2percept` Jupyter Notebook
- [X] Add `video2percept` Jupyter Notebook


arokem:
Another potentially interesting tool for video (designed for real-time
processing): https://pypi.python.org/pypi/GIFT-Grab

pulse2percept/retina.py
```diff
@@ -228,8 +228,8 @@ def __init__(self, tsample=0.005 / 1000,
         tsample : float
             Sampling time step (seconds). Default: 5e-6 s.
         tau_gcl : float
-            Time decay constant for the fast leaky integrater of the nerve
-            fiber layer (NFL); i.e., ganglion cell layer.
+            Time decay constant for the fast leaky integrater of the optic
+            fiber layer (OFL); i.e., ganglion cell layer.
```
arokem: This is confusing. Aren't OFL and GCL were two different things?

pulse2percept/retina.py
```diff
@@ -283,11 +283,11 @@ def __init__(self, tsample=0.005 / 1000,
         # Gamma functions used as convolution kernels do not depend on input
         # data, hence can be calculated once, then re-used (trade off memory
         # for speed).
-        # gamma_nfl and gamma_inl are used to calculate the fast response in
+        # gamma_ofl and gamma_inl are used to calculate the fast response in
         # bipolar and ganglion cells respectively
 
         _, self.gamma_inl = utils.gamma(1, self.tau_inl, self.tsample)
-        _, self.gamma_nfl = utils.gamma(1, self.tau_gcl, self.tsample)
+        _, self.gamma_ofl = utils.gamma(1, self.tau_gcl, self.tsample)
```
arokem: Again, this mixes OFL and GCL

pulse2percept/retina.py
```diff
@@ -472,36 +469,72 @@ def calc_per_electrode(self, pt):
...
-        # Convert the current map to one that includes axon streaks
-        ecm = effectivecurrentmap(ecs_item, pt, self.tsample)
-
-        # Run the model cascade
-        sr = self.model_cascade(ecm, layers, dojit=dojit)
+    def calc_per_pixel(self, ecs_item, pt_list, layers, dojit=False):
+        """Calculations to be done on a per-pixel basis
 
-        return sr
+        The following calculations need to be done on a pixel-by-pixel basis.
```
arokem: This repeats what was said right above. I propose to change the first line of the docstring to say what these calculations represent.

pulse2percept/retina.py
```diff
@@ -472,36 +469,72 @@ def calc_per_electrode(self, pt):
...
-def effectivecurrentmap(ecs_item, ptrain_data, tsample):
-    """
+            List of retinal layers to simulate. Choose from:
+            - 'OFL': optic fiber layer
+            - 'GCL': ganglion cell layer
+            - 'INL': inner nuclear layer
 
-    Effective current map from the electrodes in one spatial location
-    ([x, y] index) and the stimuli through these electrodes.
+        dojit : bool
```
arokem: What is this parameter?

Also: please add a "Returns" section.

pulse2percept/stimuli.py
```diff
@@ -104,14 +116,120 @@ def __init__(self, ptype, pdur, tsample, interphase_dur=0):
...
+    def __init__(self, tsample, freq=20, amp=20, dur=0.5, delay=0,
+                 pulse_dur=0.45 / 1000, interphase_dur=0.45 / 1000,
+                 pulsetype='cathodicfirst',
+                 pulseorder='pulsefirst'):
+        """A train of biphasic pulses
+
+        tsample : float
+            Sampling interval in seconds parameters, use TemporalModel.tsample.
+        ----------
+        optional parameters
```
arokem: This docstring looks a bit weird. Could you please change this into a "Parameters" block, and indicate all the optional parameters, and their default values?

pulse2percept/stimuli.py
```diff
@@ -104,14 +116,120 @@ def __init__(self, ptype, pdur, tsample, interphase_dur=0):
...
+            'cathodicfirst' has the negative phase first.
+        pulseorder : string
+            Pulse order {"gapfirst" | "pulsefirst"}, where
+            'pulsefirst' has the pulse first, followed by the gap.
+        """
+        if tsample <= 0:
+            raise ValueError("tsample must be a non-negative float.")
+
+        # Stimulus size given by `dur`
+        stim_size = int(np.round(1.0 * dur / tsample))
```
arokem: Maybe `float(dur)` to be explicit?

pulse2percept/stimuli.py
```diff
@@ -104,14 +116,120 @@ def __init__(self, ptype, pdur, tsample, interphase_dur=0):
...
+        if envelope_size > stim_size:
+            debug_s = ("Envelope size (%d) clipped to "
+                       "stimulus size (%d) for freq=%f" % (envelope_size,
+                                                           stim_size,
+                                                           freq))
+            logging.getLogger(__name__).debug(debug_s)
+            envelope_size = stim_size
+
+        # Delay given by `delay`
+        delay_size = int(np.round(1.0 * delay / tsample))
```
arokem: Same here (`float(delay)`)

pulse2percept/stimuli.py
```diff
@@ -173,81 +275,75 @@ def image2pulsetrain(img, implant, coding='amplitude', valrange=[0, 50],
...
+        # unnecessarily slow down `video2pulsetrain`.
+        raise ImportError("You do not have scikit-image installed. "
+                          "You can install it via $ pip install scikit-image.")
 
     # Make sure range of values is valid
     assert len(valrange) == 2 and valrange[1] > valrange[0]
 
-    if not isinstance(implant, implants.ElectrodeArray):
-        raise TypeError("implant must be of type implants.ElectrodeArray.")
+    isargus1 = isinstance(implant, implants.ArgusI)
```
arokem: Maybe keep an `allowed_implants` variable around (as a module-wide constant?), and check that the implant is one of them?

pulse2percept/stimuli.py
```diff
@@ -173,81 +275,75 @@ def image2pulsetrain(img, implant, coding='amplitude', valrange=[0, 50],
...
     # Make sure all pixels are between 0 and 1
     if img_orig.max() > 1.0:
         img_orig /= 255.0
-    assert np.all(img_orig >= 0.0) and np.all(img_orig <= 1.0)
 
+    # Let Scikit-Image do the resampling: Downscale image to fit array:
+    if isargus1:
+        img_stim = sit.resize(img_orig, (4, 4))
+    elif isargus2:
+        img_stim = sit.resize(img_orig, (6, 10))
```
arokem: Do the implants know about their shape? Maybe this could be derived from their properties, instead of hard-coded?

pulse2percept/tests/test_api.py
```diff
@@ -22,7 +22,7 @@ def test_Simulation_pulse2percept():
         sim.pulse2percept(pt)
 
     pt = p2p.stimuli.BiphasicPulse("cathodicfirst", 0.1, 0.005 / 1000)
-    sim.pulse2percept(pt)
+    sim.pulse2percept(pt, layers=['GCL', 'invalid'])
```
arokem: This doesn't raise an error?

pulse2percept/tests/test_files.py
```diff
@@ -3,11 +3,197 @@
...
+    npt.assert_equal(isinstance(tsout, utils.TimeSeries), True)
+    npt.assert_almost_equal(tsout.tsample, tsampleout)
+
+    # Also verify the actual data
+    tsres = tsin.resample(tsampleout)
+    tsin = None
+    npt.assert_equal(tsout.shape, tsres.shape)
+    npt.assert_almost_equal(tsout.data / 255.0, tsres.data / tsres.data.max(),
+                            decimal=0)
+    tsres = None
```
arokem: Why does `tsres` need to be set to `None` here?

pulse2percept/utils.py
```diff
@@ -113,6 +130,47 @@ def __init__(self, tsample, data):
     def __getitem__(self, y):
         return TimeSeries(self.tsample, self.data[y])
 
+    def __add__(self, other):
```
arokem: This is confusing, because I would expect an addition of the underlying arrays (element-by-element). Can we implement this as another method, not overloading `add`?

pulse2percept/__init__.py
```diff
@@ -1,29 +1,32 @@
...
                     filename='debug.log',
                     filemode='w')
-logging.getLogger(__name__).info('Welcome to pulse2percept')
 
-# Add streaming to console
+# Add streaming to console: Temporarily set level to ERROR so the imports
+# don't trigger 'deprecated' warnings
 console = logging.StreamHandler()
-console.setLevel(logging.INFO)
+console.setLevel(logging.ERROR)
```
mbeyeler: Will do the re-render.

The `deprecated` decorator is automatically executed upon module import. This is just [how decorators work](http://stackoverflow.com/questions/18062443/python-decorator-function-called-at-compile-time). An unfortunate side effect is that all these warnings are generated...
An easy work-around is to temporarily set the logging level higher, so the warnings will not print to console. There are [more sophisticated solutions](http://stackoverflow.com/questions/19601468/how-to-make-python-decorator-not-run-when-imported), but I feel like I've wasted enough time already trying to wrap my head around nested decorators.

pulse2percept/files.py
```diff
@@ -6,8 +6,453 @@
...
+    """Saves a video to file.
+
+    This function stores a NumPy ndarray to file using Scikit-Video.
+
+    Parameters
+    ----------
+    data : ndarray | pulse2percept.utils.TimeSeries
+        Video data as a NumPy ndarray must have dimension (T, M, N, C),
+        (T, M, N), (M, N, C), or (M, N), where T is the number of frames,
+        M is the height, N is the width, and C is the number of channels.
```
mbeyeler: I am using a [Scikit-Video utility](https://github.com/scikit-video/scikit-video/blob/master/skvideo/utils/__init__.py#L279) to interpret the input dimensions. The utility checks the last dimension: If it is small (<=4), it's most likely a color channel, hence the input is interpreted as (M, N, C). Else it is interpreted as (T, M, N).

I have updated the doc strings to make this clear.

pulse2percept/files.py
```diff
@@ -6,8 +6,453 @@
...
+    d_s = "Video scaled to (M, N, T) = (%d, %d, %d)." % (height, width, length)
+    logging.getLogger(__name__).debug(d_s)
+
+    # Reshape and scale the data
+    savedata = np.zeros((length, height, width, 3), dtype=np.float32)
+    for i in range(length):
+        if is_ndarray:
+            frame = skimage.img_as_float(data[i, ...])
+        elif is_timeseries:
+            frame = data.data[..., i] / data.data.max()
```
mbeyeler: It might - good catch!

pulse2percept/retina.py
mbeyeler: It should be `gamma_gcl`, and references to OFL should be replaced by GCL...

pulse2percept/stimuli.py
mbeyeler: Addressing this and the comment above: In the future, I think it would make sense to do a coordinate transform (warp) from image coordinates to locations on the retina (based on where each electrode sits). This should be able to handle arbitrarily-shaped arrays.
Currently we do not know about the an implant's shape, and the general case could be more complicated (e.g., BionicVision Australia has diamond-shaped arrangements).

Thus for the time being, I thought it to be simpler to stick to the two most common used-cases: The Argus arrays.

pulse2percept/tests/test_files.py
mbeyeler: Currently it produces a warning and ignores the layer. But I can upgrade it to an error.

pulse2percept/utils.py
mbeyeler: Sure, refactor into `append`.

pulse2percept/tests/test_api.py
mbeyeler: Currently it produces a warning and ignores the layer. But I can upgrade it to an error.

pulse2percept/tests/test_files.py
mbeyeler: Trying to save memory, but it might not be necessary anymore.

pulse2percept/files.py
```diff
@@ -6,8 +6,453 @@
...
+    else:
+        # Return as ndarray
+        return video
+
+
+def load_video_generator(filename, ffmpeg_path=None, libav_path=None):
+    """Returns a generator that can load a video from file frame-by-frame.
+
+    This function loads a video from file and returns it as a NumPy ndarray
+    with the help of the Scikit-Video.
```
mbeyeler: cut-and-paste error...

PR #75, opened 2017-06-26
[REVIEW] Changes requested from SciPy reviews

mbeyeler:
SciPy reviewers requested the following changes to the code:

- [X] Rename `dojit` to `use_jit`. Apply it only when both `has_jit` and `use_jit` are True.

The following changes will make the code equivalent to what is present in the SciPy paper, for a hypothetical 0.2 release.

- [X] Default value for `as_timeseries` in `p2p.files.load_video` should be True.
- [X] Add Horsager model.

### DIFF
--- a/pulse2percept/api.py
+++ b/pulse2percept/api.py
@@ -137,44 +137,9 @@ class Simulation(object):
         Parameters
         ----------
         model : str|retina.BaseModel
-            A custom ganglion cell model can be specified by passing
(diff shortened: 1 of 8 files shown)


arokem:
LGTM. One way to deprecate this is to keep both kwargs, where either `dojit` and `use_jit` could be used (but not both!) for now, and one raises a warning if used. But maybe let's not worry about that 😄

arokem:
Let's release! 🎉

PR #76, opened 2017-06-26
[MERGE] Improved deprecation warnings

mbeyeler:
Improved deprecation warning system, adapted from [scikit-image](https://github.com/scikit-image/scikit-image/blob/master/skimage/_shared/utils.py) and the [Python Wiki](http://wiki.python.org/moin/PythonDecoratorLibrary).

The new deprecation wrapper allows you to specify:
- which function to use instead (`alt_func`)
- a behavior (`'warn'` or `'raise'`)
- the release when the function will be removed (`removed_version`)

The function's doc string is automatically augmented to include a deprecation warning.

### DIFF
--- a/pulse2percept/utils.py
+++ b/pulse2percept/utils.py
@@ -54,23 +54,28 @@ class deprecated(object):
         The package version in which the deprecated function will be removed.
     """
 
-    def __init__(self, alt_func=None, behavior='warn', removed_version=None):
+    def __init__(self, alt_func=None, behavior='warn', deprecated_version=None,
+                 removed_version=None):
         self.alt_func = alt_func
         self.behavior = behavior
+        self.deprecated_version = deprecated_version
         self.removed_version = removed_version
 
     def __call__(self, func):
 
-        alt_msg = ''
+        alt_msg = ""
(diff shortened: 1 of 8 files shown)


pulse2percept/utils.py
```diff
@@ -36,61 +37,66 @@
...
-    alt_func = None
-    if callable(arg):
-        # Direct decoration
-        return wrap(arg)
-    else:
-        # Alternative function given
-        alt_func = arg
-        return wrap
+        # modify doc string to display deprecation warning
+        doc = '**Deprecated function**.' + alt_msg
```
arokem: Maybe we can add the `.. deprecated:: 0.2` messages in here as well, instead of manually in each docstring?

mbeyeler:
I have added a `deprecated_version` parameter. Now a complete sentence might read: "Function `scale` is deprecated since version 0.2, and will be removed in version 0.3. Use `p2p.stimuli.image2pulsetrain` instead."

The `.. deprecated:: 0.2` syntax does basically the same thing (doesn't even look much different in the Sphinx docs), so I don't think we'll need that anymore.

PR #77, opened 2017-07-12
mb/binder

mbeyeler:
Binder/Docker integration:

- [X] Add Dockerfile
- [X] Fix examples ipynb


arokem:
Do the warnings of the type "UserWarning: avconv/avprobe not found in path" mean that skvideo might not work?

mbeyeler:
It needs either `ffmpeg` or `avprobe` to work. Annoyingly, even if one of them is found (and skvideo should work), it still shows the warning message for the other one that wasn't found...

Issue #71, opened 2017-03-13
Refactor retina.make_axon_map


arokem:
Currently has some mysterious things going on in it. 

The plan: 

- [X] Write some tests for a minimal case.
- [X] Store some larger examples.
- [X] Refactor, making sure that both of the above still work.

PR #79, opened 2017-08-22
[MERGE] Fix axon map

mbeyeler:
This PR fixes and extends the spatial component of the model (closing issue #71).

Determining the contribution of axonal stimulation to the effective current map is now done axon-by-axon, not pixel-by-pixel, which allows the model to correctly handle multi-electrode stimulation. (Before, if a pixel got zero input current, the model predicted zero output - although the soma at that pixel might be connected to an axon elsewhere.)

When it comes to sensitivity of an individual axon segment, users can select from two functions:
- where sensitivity drops exponentially with distance from the soma
- where sensitivity is roughly the inverse of axonal stimulation thresholds reported in Jeng, Tang, Molnar, Desai, and Fried (2011). The sodium channel band shapes the response to electric stimulation in retinal ganglion cells. J Neural Eng 8 (036022).

Highlights:
- Two bugs in axon map calculations fixed.
- All spatial calculations are now parallelized (including the one-time building of the retinal grid).
- One-time setup now includes the following steps:
    - Grow a number of fiber bundles using the model by Jansonius et al. (2009).
    - Assume there is a neuron at every grid location: Use the above axon bundles to assign an axon to each neuron.
    - For every axon segment, calculate distance to soma. Snap axon locations to the grid using a nearest-neighbor tree structure for reduced workload.
- Calculating effective current for a given stimulus now involves:
   - Determine sensitivity of each axon segment according to a `sensitivity_rule` using the distance term calculated above.
   - Based on the sensitivity of each axon segment, determine the current contribution of that axon according to a `contribution_rule`.
- Functions `jansonius` and `make_axon_map` are now deprecated.
- Better docs, more tests.

P.S. The `p2p.retina.Grid` class is becoming a God object and should be refactored. Soon.


pulse2percept/api.py
```diff
@@ -54,25 +51,82 @@ def __init__(self, implant, engine='joblib', scheduler='threading',
...
+            function of distance from the soma. The following options are
+            available:
+            - 'decay':
+                Axon sensitivity decays exponentially with distance. Specify
+                `decay_const` to change the steepness of the fall-off with
+                distance.
+            - 'Jeng2011':
+                Axon sensitivity peaks near the sodium band (50um from the
+                soma), then plateaus on the distal axon at roughly half of the
+                peak sensitivity. See Figure 2 in Jeng, Tang, Molnar, Desai,
```
arokem: See https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt on how to integrate references in docstrings,.

pulse2percept/retina.py
```diff
@@ -901,20 +997,337 @@ def dva2ret(r_deg):
...
+            sensitivity across all axon segments. Specify `powermean_exp` to
+            change the exponent of the generalized (power) mean, calculated as
+            np.mean(x ** powermean_exp) ** (1.0 / powermean_exp). Default is 1,
+            which is equal to the arithmetic mean.
+    min_contribution : float, optional, default: 0.01
+        Current contributions below this value will not be counted.
+    decay_const : float, optional, default: 2.0
+        When `sensitivity_rule` is set to 'decay', specifies the decay constant
+        of the exponential fall-off.
+    powermean_exp : float, optional, default: 1.0
```
arokem: You might want to add error handling to catch the case where the user is trying to change `powermean_exp`, but the `contribution_rule` is not set to `mean`. I would do this by setting the default value of the argument to `None` and then checking if it is set to anything else on input, relative to `contribution_rule`

pulse2percept/api.py
```diff
@@ -472,15 +532,21 @@ def plot_fundus(self, stim=None, ax=None, upside_down=True, annotate=True):
         elif hasattr(ax, 'set_axis_bgcolor'):
             ax.set_axis_bgcolor('black')
 
-        # Draw axon pathways
-        ax.plot(retina.dva2ret(self.ofl.jan_x[:, ::5]),
-                retina.dva2ret(self.ofl.jan_y[:, ::5]),
-                c=(0.5, 1, 0.5))
+        # Draw axon pathways: Need to regenerate because parfor returns axons
+        # out of orders - random sample doesn't look neat. Should be fast:
```
arokem: Will this mess with us, if axons are generated in some way that uses non-default settings?

arokem: In particular, do I understand correctly that the default value for `n_axons` here is *different* from the default value used above?

pulse2percept/retina.py
```diff
@@ -901,20 +997,337 @@ def dva2ret(r_deg):
...
+
+    Parameters
+    ----------
+    axon: Nx2 array
+        A single axon, where axon[0, :] contains the (x, y) coordinates of the
+        location closest to the soma, and all subsequent rows move the axon
+        away from the soma towards the optic disc.
+    xg, yg: array
+        meshgrid of pixel locations in units of visual angle sp
+    tree : spat.cKDTree, optional, default: train on `xg`, `yg` meshgrid
```
arokem: => `spat.cKDTree class instance`

pulse2percept/retina.py
```diff
@@ -27,158 +33,246 @@ def __init__(self, xlo=-1000, xhi=1000, ylo=-1000, yhi=1000, eye='RE',
...
-        self.sampling = sampling
-        self.axon_id = ax_id
-        self.axon_weight = ax_wt
-        self.jan_x = jan_x
-        self.jan_y = jan_y
-        self.range_x = self.gridx.max() - self.gridx.min()
-        self.range_y = self.gridy.max() - self.gridy.min()
-
-    def current2effectivecurrent(self, cs):
+                six.moves.cPickle.dump(grid_dict, open(filename, 'wb'))
```
arokem: Nice!

pulse2percept/retina.py
```diff
@@ -27,158 +33,246 @@ def __init__(self, xlo=-1000, xhi=1000, ylo=-1000, yhi=1000, eye='RE',
...
-        for id in range(0, len(cs.flat)):
-            ecs.flat[id] = np.dot(cs.flat[self.axon_id[id]],
-                                  self.axon_weight[id])
-
-        # normalize so the response under the electrode in the ecs map
-        # is equal to cs
-        scale = np.max(cs) / ecs.flat[np.argmax(cs)]
-        ecs = ecs * scale
-
-        # this normalization is based on unit current on the retina producing
```
arokem: Do we want to retain some of this comment?

pulse2percept/retina.py
```diff
@@ -870,7 +966,7 @@ def ret2dva(r_um):
 def micron2deg(micron):
     """Transforms a distance from microns to degrees
 
-    Based on http://retina.anatomy.upenn.edu/~rob/lance/units_space.html
+    Based on http: // retina.anatomy.upenn.edu / ~rob / lance / units_space.html
```
arokem: Is this spacing intentional?

pulse2percept/retina.py
```diff
@@ -901,20 +997,337 @@ def dva2ret(r_deg):
...
 
 
+def jansonius2009(phi0, n_rho=801, rho_range=(4.0, 45.0),
+                  loc_od=(15.0, 2.0), beta_sup=-1.9, beta_inf=0.5):
+    """Grows a single axon bundle based on the model by Jansonius et al. (2009)
+
+    This function generates the trajectory of a single nerve fiber bundle
+    based on the mathematical model described in:
+
+    > Jansionus et al. (2009). A mathematical description of nerve fiber
```
arokem: Again : you might consider using the numpy docstring format for references.

pulse2percept/retina.py
```diff
@@ -1001,63 +1425,53 @@ def jansonius(num_cells=500, num_samples=801, center=np.array([15, 2]),
...
-        axon_id = axon_id + ([px],)
-
-        # now loop back along this nearest axon toward the optic disc
+        this_id = [idx]
+        this_weight = [1.0]
         for ax_pos_id in range(ax_pos_id0 - 1, -1, -1):
             # increment the distance from the starting point
-            ax = (jan_x[ax_pos_id + 1, ax_num] - jan_x[ax_pos_id, ax_num])**2
-            ay = (jan_y[ax_pos_id + 1, ax_num] - jan_y[ax_pos_id, ax_num])**2
+            ax = (jan_x[ax_pos_id + 1, ax_num] - jan_x[ax_pos_id, ax_num])
```
arokem: This fixes the bug. Could we add a test that checks this, and prevents us from ever putting it back in? (maybe you have that? I couldn't tell...)

pulse2percept/api.py
mbeyeler: Good point - I was able to pass all keyword arguments from `sim.ofl` into this code bit. The only difference is a smaller `n_axons`, so that the plot is not just all green.

pulse2percept/retina.py
mbeyeler: I did not test it in `make_axon_map`, which we now declare deprecated. I do have a test that makes sure the distance calculation is right: see `test_retina.py:test_axon_dist_from_soma`.

pulse2percept/retina.py
mbeyeler: On L.296 we say "Normalize so that the max of `ecs` is the same as `current_spread`"

arokem: Sure. It just seemed like the old comment contained a bit more information than that.

PR #81, opened 2017-10-17
[REVIEW] Adjust retinal coordinate system for left-eye implants

mbeyeler:
This pull request takes care of the difference between left-eye vs. right-eye implants: In the left eye, the coordinate system of the retina should be flipped (temporal <-> nasal) so that the OD sits at (-15, 2) deg. Implants typically are inserted from the opposite side, thus the order of electrodes changes.
In addition, this is now displayed correctly in the `plot_fundus` method.

left-right-eye


pulse2percept/implants.py
```diff
@@ -630,13 +683,24 @@ def __init__(self, x_center=0, y_center=0, h=0, rot=0 * np.pi / 180):
...
 
         # Rotation matrix
         R = np.array([np.cos(rot), np.sin(rot),
                       -np.sin(rot), np.cos(rot)]).reshape((2, 2))
 
+        # Set the x, y location of the tack
+        if self.eye == 'RE':
+            self.tack = np.matmul(R, [-5.5 * e_spacing, 0])
+        else:
+            self.tack = np.matmul(R, [5.5 * e_spacing, 0])
```
arokem: What do these constants: 5.5 here and 2.5 above, signify?

pulse2percept/tests/test_implants.py
```diff
@@ -184,6 +201,32 @@ def test_ArgusI():
...
+    npt.assert_almost_equal(argus_re.tack[1], yc)
+
+    # Left-eye implant:
+    argus_le = p2p.implants.ArgusI(eye='LE', x_center=xc, y_center=yc)
+    npt.assert_equal(argus_le['A1'].x_center > argus_le['D1'].x_center, True)
+    npt.assert_almost_equal(argus_le['D1'].y_center, argus_le['A1'].y_center)
+    npt.assert_equal(argus_le.tack[0] > argus_le['A1'].x_center, True)
+    npt.assert_almost_equal(argus_le.tack[1], yc)
+
+    # In both left and right eyes, rotat with positive angle should be CCW
```
arokem: typo: "rotat" => "rotation"

mbeyeler:
To address your comment from above: I realized flipping doesn't make sense - I believe electrodes protrude only on one side (the other is insulated). Instead the array is just rotated (as shown in the plot above).

PR #82, opened 2018-01-08
Don't require ``set_optic_fiber_layer`` to be called before ``plot_fundus``

mbeyeler:
Sometimes you just want to plot something real quick, but ``plot_fundus`` currently requires ``set_optic_fiber_layer`` to be run first - which can take a while, and is not really needed anyway. This PR removes that requirement.

### DIFF
--- a/pulse2percept/api.py
+++ b/pulse2percept/api.py
@@ -524,9 +524,24 @@ class Simulation(object):
         -------
         Returns a handle to the created figure (`fig`) and axes element (`ax`).
         """
-        from matplotlib import patches
-
-        self._set_layers()
+        if self.ofl is not None:
+            # Optic fiber layer already set via  ``set_optic_fiber_layer``
+            phi_range = self.ofl.phi_range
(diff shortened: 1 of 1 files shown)


pulse2percept/api.py
```diff
@@ -524,9 +524,24 @@ def plot_fundus(self, stim=None, ax=None, n_axons=100, upside_down=False,
...
+            phi_range = self.ofl.phi_range
+            n_rho = self.ofl.n_rho
+            rho_range = self.ofl.rho_range
+            loc_od = self.ofl.loc_od
+            x_range = self.ofl.x_range
+            y_range = self.ofl.y_range
+            plot_patch = True
+        else:
+            msg = "Optic fiber layer not set (call ``set_optic_fiber_layer``)."
+            msg += " Using default values."
```
arokem: If you want even more flexibility, you can add a `**kwargs` to the function signature and over-ride any of these default values with an input.

PR #83, opened 2018-03-26
[MRG] Cython implementation of Nanduri et al. (2012)

mbeyeler:
Single core benchmarks:
- `stim = p2p.stimuli.PulseTrain(0.05 / 1000, freq=20, amp=150, pulse_dur=0.45 / 1000, dur=0.5)`
- Convolutions (pure Python): 594 ms ± 6.72 ms
- Finite difference model (pure Python): 644 ms ± 13.8 ms
- Finite difference model (naive Cython): 166 ms ± 2.06 ms
- Finite difference model (pulse2percept): 2.96 ms ± 9.42 µs


arokem:
Looks like it might need a rebase?

pulse2percept/api.py
```diff
@@ -85,9 +85,9 @@ def set_optic_fiber_layer(self, sampling=100, x_range=None, y_range=None,
         rho_range: (rho_min, rho_max), optional, default: (4.0, 45.0)
             Lower and upper bounds for the radial position values(polar
             coordinates).
-        loc_od: (x_od, y_od), optional, default: (15.0, 2.0)
-            Location of the center of the optic disc(x, y) in Cartesian
-            coordinates.
+        loc_od: (x_od, y_od), optional, default: (15.05 1.5)
```
arokem: This should be `15.5`?

pulse2percept/fast_retina.pyx
```diff
@@ -0,0 +1,56 @@
...
+    return 1.0 / (1.0 + expf(x))
+
+cdef inline float float_max(float a, float b):
+    return a if a >= b else b
+
+DTYPE = np.float
+ctypedef np.float_t DTYPE_t
+
+
+def nanduri2012_model_cascade(np.ndarray[DTYPE_t, ndim=1] stimdata,
```
arokem: I believe that both the input to this function (`stimdata`) as well as the output (`out_R4`) can be implemented as memory views. You might also get something out of refactoring the intermediate list (`tmp_R4a`) as a memory view into an array. Jake's excellent blogpost on the topic (https://jakevdp.github.io/blog/2012/08/08/memoryview-benchmarks/) explains why: memory views are designed to reduce overhead in reference counting when slicing into arrays (IIUC).

mbeyeler:
There's more we could do in terms of threading and parallelization, but I think it's good for now (`0.0-usage-example.ipynb` now runs in 30 secs instead of 20 mins).

arokem:
👏 merging!

PR #84, opened 2018-08-15
Add visualization module

mbeyeler:
- move `plot_fundus` to visualization module
- fix nasal/temporal labeling for left eyes
- turn annotation on/off


arokem:
Yep. Makes sense!

Issue #90, opened 2019-04-12
Inconsistent tsample values break simulation


mbeyeler:
This simple example:
```
import pulse2percept as p2p
sim = p2p.Simulation(p2p.implants.ArgusII())
pulse= {'F2': p2p.stimuli.PulseTrain(0.01/1000)}
percept = sim.pulse2percept(pulse)
```
breaks during the last line with "For now, all pulse trains must have the same sampling time step as the ganglion cell layer. In the future, this requirement might be relaxed."
(The problem is that the ganglion cell layer uses tsample=0.005/1000 per default. And PulseTrain doesn't have such a default value.)

First of all, this error message is very confusing. As a user I've never even heard of a ganglion cell layer. Let alone touched it.

Second, if every method must use the same `tsample`, specify it once at the global level (maybe in `p2p.Simulation` constructor?). Otherwise at least have all methods use the same default value for `tsample`...

mbeyeler:
Closed via #96 and #119

PR #119, opened 2019-08-09
Add Stimulus object

mbeyeler:
- [X] defines a common interface for all stimuli
- [X] can handle stimuli with and without a time component
- [X] creates stimuli from various source types (e.g., scalars, NumPy arrays, `TimeSeries`, lists, dicts)
- [X] efficiently stores `TimeSeries` data (dropping redundant items)
- [X] removes `xarray` dependency
- [X] automatically fixes issue with inconsistent tsamples (issue #90)

In the future, this might replace `TimeSeries`.

### DIFF
--- a/pulse2percept/stimuli/base.py
+++ b/pulse2percept/stimuli/base.py
@@ -2,7 +2,7 @@ import numpy as np
 np.set_printoptions(precision=2, threshold=5, edgeitems=2)
 from copy import deepcopy as cp
 from collections import OrderedDict as ODict
-from scipy import interpolate as spi
+from scipy.interpolate import interp1d
 
 from .pulse_trains import TimeSeries
 from ..utils import PrettyPrint
@@ -20,15 +20,6 @@ class Stimulus(PrettyPrint):
     A stimulus can be created from a variety of source types (see below),
     including lists and dictionaries. Depending on the source type, a stimulus
     might have a time component or not.
-
-    .. note::
-       A compression method is used to `sparsify` large and redundant data
-       sources such as `TimeSeries`.
-
-       This also means that all-zero source data and electrodes with all-zero
-       activation will be trimmed from the stimulus.
-
-       Set `sparsify=False` to deactivate this behavior.
 
     Parameters
     ----------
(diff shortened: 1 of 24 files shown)


pulse2percept/stimuli/base.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        if not isinstance(other, Stimulus):
+            return False
+        if self.time is None:
+            if other.time is not None:
+                return False
+        else:
+            if other.time is None:
+                return False
+            if len(self.time) != len(other.time):
```
arokem: This means that two stimuli would be unidentical if one is a "compressed" version of the other. Meaning that you could have: 

```
>>> s1 == s2 
False
>>> s1.compress()
>>>s1 == s2 
True
```

That seems fine to me, but want to make sure that's intended.

pulse2percept/stimuli/base.py
```diff
@@ -0,0 +1,465 @@
...
+        Compare two Stimulus objects:
+
+        >>> from pulse2percept.stimuli import Stimulus
+        >>> stim1 = Stimulus(np.ones(3))
+        >>> stim2 = Stimulus(np.zeros(5))
+        >>> stim1 != stim2
+        True
+
+        """
+        return not self.__eq__(other)
```
arokem: Would be neat to implement the `__get_item__` that takes a time point and returns the data at that time point, including interpolating as needed. But maybe on a separate PR.

pulse2percept/stimuli/tests/test_base.py
```diff
@@ -0,0 +1,239 @@
...
+
+def test_Stimulus___eq__():
+    # Two Stimulus objects created from the same source data are considered
+    # equal:
+    for source in [3, [], np.ones(3), [3, 4, 5], np.ones((3, 6))]:
+        npt.assert_equal(Stimulus(source) == Stimulus(source), True)
+    stim = Stimulus(np.ones((2, 3)), compress=True)
+    # Compressed vs uncompressed:
+    npt.assert_equal(stim == Stimulus(np.ones((2, 3)), compress=False), False)
+    npt.assert_equal(stim != Stimulus(np.ones((2, 3)), compress=False), True)
```
arokem: OK - this asserts my comment on the code, meaning that it is intended.

setup.py
```diff
@@ -22,7 +22,6 @@
 SCIPY_MIN_VERSION = '1.0'
 CYTHON_MIN_VERSION = '0.28'
 JOBLIB_MIN_VERSION = '0.11'
-XARRAY_MIN_VERSION = '0.12'
```
arokem: 😞

pulse2percept/stimuli/base.py
mbeyeler: I added this feature request in issue #121

Issue #122, opened 2019-08-25
Specify electrode type for ElectrodeGrid


mbeyeler:
`ElectrodeGrid` currently assumes all electrodes are `DiskElectrode` objects.

To be more flexible, constructor needs to accept an electrode type. Thing is, `DiskElectrode` needs a radius, but `PointSource` does not. Therefore, we also need a way to specify optional keyword arguments that can be passed to the electrode constructor.

arokem:
Add a ‘**elec_kwargs’ input to the grid constructor?

mbeyeler:
Closed via #150

PR #150, opened 2020-02-10
[ENH] update the electrode_type for ElectrodeGrid

DylanLin29:
## Description

- [X] Specify electrode type for `ElectrodeGrid` (see issue #122)

## Checklist

- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./examples` (new features only)
- [ ] Unit tests
- [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)


pulse2percept/implants/base.py
```diff
@@ -347,18 +347,20 @@ class ElectrodeGrid(ElectrodeArray):
     * Allow user-specified electrode types (see :issue:`122`).
 
     """
-
-    def __init__(self, shape, x=0, y=0, z=0, rot=0, r=10, spacing=None,
-                 names=('A', '1')):
+    
+    def __init__(self, shape, x=0, y=0, z=0, rot=0, r = 10, spacing=None,
```
mbeyeler: You are on the right track. Because different electrode objects have different number of parameters, it would be better change the function signature to:

```
def __init__(self, shape, x=0, y=0, z=0, rot=0, spacing=None,
                  electrode_type='DiskElectrode', electrode_kwargs=None):
```

where `electrode_kwargs` is a dictionary with optional keyword arguments for `electrode_type`.

So the user could do:
- `ElectrodeGrid((4, 4), electrode_type='PointSource')
- `ElectrodeGrid((4, 4), electrode_type='DiskElectrode', electrode_kwargs={'r': 10})`

This way, we don't need to change the function signature when we add more electrode types in the future.

pulse2percept/implants/base.py
```diff
@@ -347,18 +347,20 @@ class ElectrodeGrid(ElectrodeArray):
...
         if not isinstance(names, (tuple, list, np.ndarray)):
             raise TypeError("'names' must be a tuple/list of (rows, cols)")
-
         if not isinstance(shape, (tuple, list, np.ndarray)):
             raise TypeError("'shape' must be a tuple/list of (rows, cols)")
         if len(shape) != 2:
             raise ValueError("'shape' must have two elements: (rows, cols)")
         if np.prod(shape) <= 0:
             raise ValueError("Grid must have all non-zero rows and columns.")
+        if electrode_type != 'DiskElectrode' and electrode_type != "PointSource": # check the value of the electrode_type
```
mbeyeler: - If type is 'DiskElectrode': make sure `electrode_kwargs` contains an entry for `r` (which is the radius)
- If type is 'PointSource': do nothing
- else: `raise NotImplementedError`

pulse2percept/implants/base.py
```diff
@@ -504,9 +508,14 @@ def _set_grid(self):
         x_arr += self.x
         y_arr += self.y
 
-        # TODO parameterize the type of electrode
-        for x, y, z, r, name in zip(x_arr, y_arr, z_arr, r_arr, names):
-            self.add_electrode(name, DiskElectrode(x, y, z, r))
+        
+        if self.electrode_type == 'DiskElectrode':
+            for x, y, z, r, name in zip(x_arr, y_arr, z_arr, r_arr, names):
+                self.add_electrode(name, DiskElectrode(x, y, z, r))
```
mbeyeler: Here you want to grab the radius from `electrode_kwargs['r']`

pulse2percept/implants/base.py
```diff
@@ -504,9 +508,14 @@ def _set_grid(self):
...
+        # when the electrode type is PointSource
+        else:
```
mbeyeler: make it more explicit: `else if self.electrode_type == 'PointSource'`

PR #187, opened 2020-05-15
[ENH] Prima implant

DylanLin29:
## Checklist

- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./examples` (new features only)
- [ ] Unit tests
- [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)


pulse2percept/implants/base.py
```diff
@@ -520,26 +524,51 @@ def _make_grid(self, spacing, x, y, z, rot, names, etype, **kwargs):
...
-            x_arr = []
-            for row in range(rows):
-                if row % 2:
-                    x_arr.append(x_arr_rshift[row])
-                else:
-                    x_arr.append(x_arr_lshift[row])
-            x_arr = np.array(x_arr)
-            y_arr = y_arr_rshift
+            # When orientation is horizontal, x_arr is shifting to create the hex grid
+            if orientation == 'horizontal':
```
mbeyeler, line 528: I like the split between 'horizontal' and 'vertical', but there seems to be something wrong with the grid:

image

You can see that the distance between columns is too large. Not sure if that is a problem with both orientations. Please track down the error, and when you find it, also add a test that makes sure this issue is fixed.

pulse2percept/implants/prima.py
```diff
@@ -0,0 +1,28 @@
+"""`Prima`"""
+import numpy as np
+from collections import OrderedDict
```
mbeyeler, line 3: OrderedDict is unused

pulse2percept/implants/prima.py
```diff
@@ -0,0 +1,28 @@
...
+from pulse2percept.implants import ElectrodeGrid, ProsthesisSystem, DiskElectrode
+
+class Prima(ProsthesisSystem):
```
mbeyeler, line 6: This is fine, but don't forget to also add it to `__init__.py` so it can be imported

Issue #189, opened 2020-05-28
[BUG] Special characters in ElectrodeGrid names


mbeyeler:
If an `ElectrodeGrid` has more than 26 rows/columns, the row/column names become unusable, because they include special characters like `\\` and `~`. Instead, make it so the names go `A-Z`, then `AA-AZ`, then `BA-BZ`, etc.

mbeyeler:
Closed via #192

PR #192, opened 2020-05-28
[FIX] ElectrodeGrid naming convention

mbeyeler:
Fixes `ElectrodeGrid` naming convention when there are more than 26 alphabetic rows/columns (issue #189): new ordering is `A-Z`, then `AA-AZ`, then `BA-BZ`, etc.


mbeyeler:
I see, wheels are failing because https://status.quay.io/ is down

PR #306, opened 2020-09-25
[DOC] Update ImageStimulus docstring - Closes Issue #252

### LINKED ISSUE #252: [BUG] ImageStimulus doc wrong/inconsistent in several places
Couple of small things:

* The docstring of `ImageStimulus.resize` for the `shape` parameter should say: "Shape of the resized image. If one of the dimensions is set to -1, its value will be inferred by keeping a constant aspect ratio."
* The docstring of `ImageStimulus.rgb2gray` for the return parameter should say: "A copy of the stimulus object with all RGB values converted to grayscale in the range [0, 1]."

Conrad-Crowley:
## Description
Corrects inconsistencies in the ImageStimulus Class' docstring raised in #252 . 
Closes #252 (added to link issue)

### DIFF
--- a/pulse2percept/stimuli/images.py
+++ b/pulse2percept/stimuli/images.py
@@ -49,9 +49,8 @@ class ImageStimulus(Stimulus):
         https://imageio.readthedocs.io/en/stable/formats.html.
 
     resize : (height, width) or None, optional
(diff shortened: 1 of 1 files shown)


mbeyeler (approved):
LGTM! Thanks

Issue #318, opened 2020-10-27
[BUG] Horsager fixed duration stimuli should have interphase duration of 0 


jgranley:
interphase_dur should be 0 for Horsager 2009 fixed pulse duration data

mbeyeler:
Note to future self: This is an assumption on our part, because it is not mentioned in the paper, but switching to `interphase_dur=0` greatly reduces the model fit error. Should clarify this in the dataset description

Issue #237, opened 2020-07-21
Horsager 2009 - Mixed-up between suprathreshold and threshold experiment of bursting triplets [BUG]


khahmnguyen:
**Describe the bug**
Labels of bursting triplets experiments are mixed up: 
```
data.stim_type == 'bursting_triplets' should be from figs3.6
data.stim_type == 'bursting_triplets_supra' should be from figs 7
```

PR #329, opened 2020-11-16
[REF] Updated horsager data 

jgranley:
This branch includes various updates to the horsager2009.csv dataset:

- Fixed duration stimuli now have interphase duration of 0 (#318)

- Bursting triplets and bursting triplets supra data have been 'unswapped' (#237). Previously the data for bursting triplets was actually the data for bursting triplets supra, vice versa, and multiple other columns such as reference stimuli and theta values were mixed up between the two. Now all of the data is correct has the appropriate label.

- For variable duration, missing data points have been added and interpolation has been removed, since we do not actually know the exact frequencies the original experiment used.

- Horsager test cases have been updated to reflect the new data. No new test cases were added, but the values of the original test cases were updated to account for the fact that we now have more data points.


mbeyeler (approved):
Looks good to me. Well done!

PR #331, opened 2020-12-02
[ENH] Add KDTree to find closest axon in AxonMapModel

jgranley:
## Description
When building the axon map model, a KDTree is now used to find the nearest axon to each simulated point, instead of searching over each axon segment. This changes closest axon lookup time to scale logarithmically in number of axon segments instead of linearly, resulting in speedups of around 150-200x (for closest axon lookup, not for build) using default parameters. The remaining bottleneck in building the model is calculating axon contribution.

Using KDTree is now the default. This can be overridden by passing use_legacy_build=True in the constructor, which uses the old method of searching over all axons. 

The default number of axons is now set to 1000. This could easily be set much higher without significant effect on build time. 

Full list of changes:
- Added [cKDTree ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html) from scipy as default to find closest axons 
- Added use_legacy_build parameter to AxonMapModel and AxonMapSpatial, with docstring
- Set default n_axons=1000
- Added test cases for use_legacy_model = True and False
- Refactored predict_percept test cases to no longer rely on the default n_axons to be 500
- Fixed docstrings for xrange and yrange (docstring and parameter name didn't match before)

Notes:
The tree is built inside of find_closest_axon and discarded after use because of the simpler implementation, and relatively quick build time. This means that if find_closest_axon() is repeatedly called on the same set of bundles, performance may suffer. During the AxonMapModel build, find_closest_axon() is only called once, so this is not a problem. If the model is ever changed to call find_closest_axon() again postbuild, it would be better to build the KDTree in _build() instead and store it for later use.

### DIFF
--- a/pulse2percept/models/beyeler2019.py
+++ b/pulse2percept/models/beyeler2019.py
@@ -206,7 +206,7 @@ class AxonMapSpatial(SpatialModel):
             'axlambda': 200,
             # Set the (x,y) location of the optic disc:
             'loc_od': (15.5, 1.5),
-            'n_axons': 500,
+            'n_axons': 1000,
             'axons_range': (-180, 180),
             # Number of sampling points along the radial axis (polar coords):
             'n_ax_segments': 500,
@@ -641,12 +641,12 @@ class AxonMapModel(Model):
         Exponential decay constant away from the axon(microns).
     eye: {'RE', LE'}, optional
         Eye for which to generate the axon map.
-    x_range : (x_min, x_max), optional
+    xrange : (x_min, x_max), optional
         A tuple indicating the range of x values to simulate (in degrees of
         visual angle). In a right eye, negative x values correspond to the
         temporal retina, and positive x values to the nasal retina. In a left
         eye, the opposite is true.
-    y_range : tuple, (y_min, y_max)
+    yrange : tuple, (y_min, y_max)
         A tuple indicating the range of y values to simulate (in degrees of
         visual angle). Negative y values correspond to the superior retina,
         and positive y values to the inferior retina.
@@ -675,6 +675,9 @@ class AxonMapModel(Model):
         Axon segments whose contribution to brightness is smaller than this
         value will be pruned to improve computational efficiency. Set to a
         value between 0 and 1.
+    use_legacy_build: bool, optional
+        If true, searches over axons instead of using KDTree. Build will 
+        likely be slower if True
     axon_pickle: str, optional
         File name in which to store precomputed axon maps.
     ignore_pickle: bool, optional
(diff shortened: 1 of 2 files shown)


pulse2percept/models/beyeler2019.py
```diff
@@ -378,16 +384,24 @@ def find_closest_axon(self, bundles, xret=None, yret=None):
...
+                closest_seg = fast_find_closest_axon(flat_bundles,
+                                                     xret.ravel(),
+                                                     yret.ravel())
+            else:
+                closest_seg = [np.argmin((flat_bundles[:, 0] - x) ** 2 +
+                                         (flat_bundles[:, 1] - y) ** 2)
+                               for x, y in zip(xret.ravel(),
+                                               yret.ravel())]
+        else: 
+            kdtree = cKDTree(flat_bundles, leafsize=60)
```
mbeyeler, line 399: Can you add a comment about how you chose `leafsize` - do you think this value will always work, or should it be something that can be set by the user?

mbeyeler (changes requested):
Great work as always! I just had a quick question about `leafsize` and whether it should be settable by the user

pulse2percept/models/beyeler2019.py
jgranley: Benchmarking the best leaf size was somewhat difficult. I experimented with leaf sizes in the range [10, 300] for number of axons in the range [500, 10000], and eventually concluded that leaf size didn't affect the result much. 

When changing leaf size for a fixed number of axons, the total fit+query time never changed by more than 200ms (max - min < 200). Further when running multiple trials of the same leaf size and num_axons, the total time changed by up to ~170ms (max-min). Thus it was hard to determine if changing leaf size was actually affecting run time, or if it was just random variance between trials. 

After running a large number of trials, it seemed like any leafsize in the range [20, 100] was essentially identical, with marginal performance drops otherwise.

In general, the optimal leaf size increased slightly with number of axons, but since changing leaf size never affected the result by more than 200ms, I didn't think it would be worth spending a lot of time figuring out what the best leaf size was for a passed number of axons, and just left it as 50.

jgranley: So in theory the current value of leafsize should work pretty well.

Letting the user change it is extremely easy though, so I can also just add it as a parameter that can optionally be passed to find_closest_axon

mbeyeler: That's ok. We don't need to make it a settable option if the behavior doesn't change much. Choosing a good value for the user simplifies things. I was just wondering if it would make a difference.

mbeyeler: We can merge this as is. Great job!

PR #336, opened 2020-12-10
[FIX] Axon streak length 

jgranley:
## Description
Previously axon streaks were way too long. This was because the distance along the axon was not being calculated correctly. Previously, the code in calculate_axon_contributions took the cumulative sum of the distance squared between each segment. Now, it sums the distances, and squares at the end. 

Test cases for calculate_axon_contribution and predict_percept were also updated


mbeyeler (approved):
Beautiful!

PR #346, opened 2021-02-03
[ENH] Stimulus metadata improvements

jgranley:
Previously, stimulus metadata was lost in various places as stimuli were created, and during percept prediction.

Now, metadata information is passed along as stimuli are created/manipulated.
If a stimulus is created from a collection of sources, the stimulus object collects the metadata from each of the sources and stores it in the corresponding electrode in stim.metadata['electrodes']

Additionally, BiphasicPulseTrain is modified to store its stimulus parameters (such as frequency, amplitude, etc) in its metadata. This ensures the information is not lost when a stimulus is created from multiple BiphasicPulseTrains, and enables extension models such as BiphasicAxonMapModel

New test cases were added to test stimulus metadata

### DIFF
--- a/doc/topics/stimuli.rst
+++ b/doc/topics/stimuli.rst
@@ -290,8 +290,7 @@ in ''metadata["electrodes"][electrode]["metadata"]''
     stim.metadata
 
     # Multiple source metadata
-    stim = Stimulus({'A1' : BiphasicPulseTrain(1,1,1, metadata='A1 metadata'), 
-                     'B2' : BiphasicPulseTrain(1,1,1, metadata='B2 metadata')},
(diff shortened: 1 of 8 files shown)


mbeyeler (changes requested):
Looks great!

Small request: Can you briefly talk about accessing metadata in the docs? In file `doc/topics/stimuli.rst`, add a new subsection "Accessing metadata" on line 270. Briefly explain how some stimulus objects store their own metadata, but that the user can provide their own. If they provide their own, it will be stored in `stim.metadata['user']` (they way I understand it)

jgranley:
Done!

mbeyeler (approved):
LGTM!

PR #347, opened 2021-02-03
[ENH] Add amplitude vs size points to Nanduri2012 dataset

jgranley:
Adds points from figure 4c in Nanduri et al 2012 to the nanduri2012.csv dataset, which describe how percept size changes with amplitude changes

load_nanduri was updated to be able to select points by task in addition to electrode (either 'rate' for brightness rating, which consists of all of the previous points, as well as 'size' which selects only the new points). 

A few column names were also slightly refactored to match the csv as generated from Data Warehouse

Test cases were updated to reflect new dataset and to test loading dataset filtering by task


mbeyeler (approved):
LGTM!

Let's wait with merging until Python 3.5 support is dropped (PR #348)

PR #354, opened 2021-02-18
[FIX] Fix percept.plot() normalization bug

jgranley:
Fixes a bug where if the user passed vmin and vmax parameters to percept.plot() they were ignored and reset to frame.min() and frame.max().


mbeyeler (changes requested):
Can you add a test in the `test_Percept_plot()` function of `percepts/tests/test_base.py` to make sure it now works? Something like:
1. call plot with vmin=2, vmax=4
2. make sure `ax.collections[0].get_clim()` is in [2, 4]

mbeyeler (approved):
LGTM!

PR #398, opened 2021-09-17
[ENH][REF] Add BiphasicAxonMapModel

jgranley:
Implements the model in "Granley, J., & Beyeler, M. (2021). A Computational Model of Phosphene Appearance for Epiretinal Prostheses. EMBC21" as BiphasicAxonMapModel

This PR includes a number of additions / changes:
- BiphasicAxonMapModel and BiphasicAxonMapSpatial, subclasses of AxonMap
- Cython implementation of predict_percept for BiphasicAxonMap
- Preliminary support for Jax engine: Model builds correctly with uniform length axons, but predict_percept is to be implemented in a future release.
- Test cases for BiphasicAxonMap
- Updated test cases for AxonMap
- Default implementation for effect models (F_streak, F_size, F_bright)
- Add new references to doc
- One more stimulus metadata improvement

BiphasicAxonMap implements the general model from the paper. What equations it uses for F_streak, F_size, and F_bright can be changed by passing a python callable with signature` f(freq, amp, pdur) -> float `to  `model.bright_model`, `model.size_model`, or `model.streak_model`. Additionally, the coefficients a0-a9 can be changed by directly through model (e.g.  `model.a0`) 

The changes to models/base.py and utils/base.py are so that BiphasicAxonMapSpatial can properly delegate setattr and getattr calls to its size / streak / brightness models.


pulse2percept/models/granley2021.py
```diff
@@ -0,0 +1,559 @@
+"""`BiphasicAxonMapModel`"""
+from attr import Attribute
+import numpy as np
+from copy import deepcopy
+import sys
+
+from numpy.core.fromnumeric import size
```
mbeyeler, line 7: Some of these imports seem to be unused, e.g. `Attribute`, `deepcopy`, `size`,  `load_nanduri2012`. Would be good to remove them to avoid unnecessary dependencies

pulse2percept/models/granley2021.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def _build(self):
+        if self.bright_model is None:
+            self.bright_model = DefaultBrightModel(
+                do_thresholding=self.do_thresholding)
+        if self.size_model is None:
+            self.size_model = DefaultSizeModel(self.rho)
+        if self.streak_model is None:
+            self.streak_model = DefaultStreakModel(self.axlambda)
+        assert(callable(self.bright_model))
```
mbeyeler, line 331: `assert` might be ignored in pip-packaged code. It's better to do `if not callable(...): raise TypeError("'bright_model' should be blah, not bleh")`

pulse2percept/models/granley2021.py
```diff
@@ -0,0 +1,559 @@
...
+    Default model to be used for brightness scaling in BiphasicAxonMapModel
+    Implements Eq 4 from [Granley2021]_ 
+    Fit using data from [Nanduri2012]_ and [Weitz2015]_
+
+    Parameters:
+    ------------
+    do_thresholding : bool, optional
+        Set to true to enable probabilistic phosphene appearance at near-threshold 
+        amplitudes
+    **kwargs : dict, optional
```
mbeyeler, line 31: In all of these models, it would be tremendously useful to spell out the meaning of the different A0-9 parameters, rather than `**kwargs`

pulse2percept/models/granley2021.py
```diff
@@ -0,0 +1,559 @@
...
+    Parameters:
+    ------------
+    do_thresholding : bool, optional
+        Set to true to enable probabilistic phosphene appearance at near-threshold 
+        amplitudes
+    **kwargs : dict, optional 
+        Allows for setting model coefficients, A0-4
+    """
+
+    def __init__(self,  do_thresholding=False, **kwargs):
```
mbeyeler, line 35: If you make this class derive from `BaseModel`, you can inherit all the default value and `setattr` logic. Then you don't need a constructor, you just need to supply a `get_default_params` method that returns a dict `{'a0': 0.27, ...}`. Deriving from `BaseModel` also gives you all the `Frozen` logic, where params can't be added outside init/build

mbeyeler: In fact, I believe this should help you solve the issue of setting parameter values before build. It would no longer be `DefaultBrightModel`, just `BrightModel` I guess. You'd call it like `BrightModel()` to use the default params, or `BrightModel(a0=1)` etc. to override stuff

pulse2percept/models/granley2021.py
```diff
@@ -0,0 +1,559 @@
...
+            A flag whether to ignore the pickle file in future calls to
+            ``model.build()``.
+    """
+
+    def __init__(self, **params):
+        super(BiphasicAxonMapSpatial, self).__init__(**params)
+
+    def __getattr__(self, attr):
+        # Called when normal get attribute fails
+        if sys._getframe(2).f_code.co_name == '__init__' or  \
```
mbeyeler, line 251: Should this be added to the `Frozen` class, or does it just apply here because you're nesting models/classes?

mbeyeler (changes requested):
I did a quick pass and had some comments. I think inheriting from `BaseModel` should help you with setting param values before build. Also some minor stuff

pulse2percept/models/granley2021.py
jgranley: It applies here because I overrided __setattr__, effectively adding one stack to the frame. Thus init appears in position 3 instead of position 2. I think it should be okay for the frozen class though, because if init is anywhere in the call stack frame, then this was still part of the constructor.

jgranley:
Updates:
- Removed unnecessary imports
- Default effects models now inherit from BaseModel
- Effects model parameters can be set before build
- Removed assert calls
- Added docstrings to explain the meaning of A0-A9 parameters

pulse2percept/models/granley2021.py
jgranley: Just to clarify, i modified the code to print out the top 5 frames when `__setattr__` is called. 
AxonMapModel:
`['set_attr', '__init__', '__init__', '__init__', '__init__']`
BiphasicAxonMapModel:
`['set_attr', '__setattr__', '__init__', '__init__', '__init__']`
Because of the way the frozen class and nested model classes work, most subclasses never call their own __setattr__, they just use the frozen one from the parent class instead. Since I need to distribute calls to `__setattr__` to the relevant effects models, BiphasicAxonMapSpatial implements its own `__setattr__`, which adds the call frame to the stack

mbeyeler: Gotcha - that makes sense

mbeyeler (approved):

PR #441, opened 2021-11-29
[ENH] Jax support for BiphasicAxonMapModel

jgranley:
## Description
This pull request adds Jax support for the BiphasicAxonMapModel (finally)

**TLDR:**
- Jax predict_percept is fast. ~100x faster on GPU and ~2x faster on CPU
- Batched version of predict_percept
- Recompilation is not yet addressed
- Memory limits computational speed for batching / NN
- Updated tests 

### Change Summary:
**Jax**
Similarly to elsewhere in p2p, jax is only [imported](https://github.com/jgranley/pulse2percept/blob/jax_biphasic/pulse2percept/models/granley2021.py#L15-L31) if it is installed, and the model works regardless of if jax is installed or not. This is what the conditional decorator is needed for.

Currently, jax is not added to requirements.txt or requirements-dev.txt, due to complex installation (for GPU version) and it is not compatible with windows. 

**Performance improvements:**
I chose to balance speed with usability / simplicity. This version is not quite as fast as the best performing one, but it is functionally just as close, and significantly simpler and more stable (not nearly as much jax black magic)

All metrics are with default settings unless otherwise specified. Also, all measure are of jax vs cython time only. Stimulus manipulation / percept creation adds to this. 

GPU:
- 1 electrode: 10ms -> 0.2ms (50x speedup)
- 15 electrode: 105ms -> 1ms (100x speedup)
- 1000 electrode: 6s -> 20ms (300x speedup)
- Batch of 256 60 electrode percepts: 1.5min -> 200ms with lax.map, 5ms with vmap (400x or 5000x speedup), see below

CPU:
- 1 electrode: 10ms -> 3.8ms (~2x speedup)
- 15 electrode: 105ms -> 36ms (~3x speedup)
- 1000 electrode: 6s -> 2s (3x speedup)

In addition, performance now scales linearly with number of electrodes, grid size, and batch size. Before, when number of electrodes / grid size increased, runtime increased quadratically or worse. This means that large implants or detailed simulations are now possible. 

**Predict Percept**
If the engine is jax, `model._predict_spatial` calls `model.predict_spatial_jax`. This function is jit'ed and supports autograd.

In order for the full model to be differentiable, the effects models (bright, size, streak) also need to implemented in jax. Effect models are getting somewhat complicated, it could be good to disentangle them from the axon map model (abstract the biphasic model completely) at some point in the future....

One issue with this implementation is that Jax is meant to be completely functional and stateless, meaning it shouldnt access member variables or globals. However, in order to integrate jax into the existing model flow, it is nearly impossible to accomplish this in any sort of clean and usable way. Thus, some of the jax functions are stateful.
When a jax function accesses a global or member variable, its possible that it will reuse a past cached result instead of actually reading the value. This also allows gradients to ignore the effect of stateful code. 

As it is currently coded, this should not be an issue unless the user changes one of these model parameters post build (rho, lambda, axon_contrib, engine, thresh_percept, or any effect model). The jax behaviour is undefined if this happens. Upon build the `static_argnums` argument to all of the jit'ed functions causes everything to be recompiled anyways.

**Batched Predict Percept**
There are also 2 new functions supporting batched percept prediction: `_predict_spatial_batched` and `predict_percept_batched`. `_predict_spatial_batched` is a jax function (fast, jit'ed, differentiable) to predict the raw percept data from arrays of stimulus parameters, while `predict_percept_batched` takes a list of stimulus objects, converts them to jax form, runs the optimized jax function, and returns a list of `Percept` objects.

There are two current limitations: 
**First**, `_predict_spatial_batched` uses `lax.map` instead of `vmap` due to GPU memory restrictions. Even on a 3090 (24gb), computing a single batch of 256 percepts with 15 stimulating electrodes and default grid size each barely fits into the memory. Any larger is infeasible. This is because when using vmap as opposed to looping, jax tries to compute everything all at once using giant matrix multiplications. With default grid size (121x121), this results in jax trying to create multiple 121x121x15x100x256 = 21GB arrays. On our 3090's and with a smaller grid size this might be manageable, but for larger number of electrodes and for the general use case this is not practical. Thus I had it use lax.map instead of vmap, which is essentially an optimized loop. This is much slower (200ms vs 4ms). 

In the future we could look into if this is the best option. It might be possible to restructure the operation to not use as much memory while keeping the speedup. But even if not, there are a number of options:
- lax.map across the batch dimension (current). This has the added benefit that it is theoretically possible to support batches with percepts of different size, with some refactoring to support nonuniform numpy arrays. Also, single percept prediction is still completely vmap-ed and fast
- use lax.map to break up vmap across simulated points, vmap across batch. I think it might be possible instead of vmapping across all  simulated points (by default 14k each with 100 axon segments), break it up into reasonable chunks using lax.map, and then vmap the rest
- A larger rework might refactor the use precomputed axon contributions. Currently, we are doing a lot of duplicate work. Under default settings, each of ~14k points has ~100 axon segments, each of which has its contribution precomputed and looped over in `predict_percept`. This is larger than the total number of axon segments even simulated (1000 * 500), so it might be possible to refactor to precompute over the entire map and then use the correct portion in predict_percept.

**Second**, all stimuli in a batch passed to  `_predict_spatial_batched`  must have the exact same electrodes. `predict_percept_batched` finds the minimal set of electrodes covering all of the stimuli passed and pads the stimulus with 0's to make it uniform before passing to the jax function, so this is not a problem for the end user use case. However this version is not differentiable and is slow. This means that any neural network using this batching needs to have the exact same set of active electrodes every time. For argus this is managable, but for larger implants this will quickly become infeasible. Using lax.map (current implementation) it might be possible to change this in the future, but there are still further complications with using variable length arrays and integrating with a neural network.

**Tests**
All current tests have been updated to also test if engine is jax, and some new tests have been added to test that jax engine is the same as cython engine, and that predict_percept_batched works as expected.
Currently, all tests pass locally under all configurations (jax cpu, jax gpu, and jax not installed). The test suite set up on github is only testing the configuration without jax being installed, so all the jax tests are just skipped. 

**Bright Model a0 set to 0**
Currently, if a pulse with freq, amp, and pdur of (0,0,0) is fed through `predict_spatial`, the brightness output is nonzero. Under normal circumstances, `predict_percept` never allows this to happen, but in jax we are no longer allowed to conditionally return 0 for the brightness contribution whenever the amp is 0 on a certain electrode. Further, batching pads stimuli with 0's to get uniform shape. 

To fix this, I simply set a4, the F_bright intercept, to 0. This logically makes sense (no amp = no percept) but results in a very slight worse fit to nanduri data. If the user changes a4 or bright_model such that this is no longer the case, then `predict_percept_batched` will no longer work as expected.

**Recompilation**
Currently, any time `predict_percept` or `predict_percept_batched` is called with a different number of electrodes, the jax function has to be recompiled. It would be possible to pre-run certain combinations in `build`, or to pad all predictions to a configurable length interval (e.g. pad to multiples of 10 electrodes, and precompile for all of those). But I chose not to because it adds a lot of complication, compilation time is not terrible (less than a second), and because `predict_percept_batched` already forces constant number of electrodes so this doesn't really affect batching. 

## Type of Change


mbeyeler (approved):
lgtm!

PR #450, opened 2022-01-05
[MNT] Switch to f-strings

### LINKED ISSUE #428: [MNT] Switch to f strings
Now that Python >= 3.6, we can use [f strings](https://realpython.com/python-f-strings/)

narenberg:
## Description

Refactor commit that changes all the %-strings and .format()-strings I could find with f-strings to get faster and keep up with new standards.

Closes #428 

## Type of Change

## Checklist

I'm not sure if the commenting, documentation, and adding tests is relevant here since I'm just changing the format of strings.


narenberg:
Alright it builds on my local machine. 2 tests fail when I run the test suite but they also fail on master so I don't think these changes cause the failures.

PR #453, opened 2022-01-12
[ENH] Let User Specify Number Of OpenMP Threads

### LINKED ISSUE #263: [ENH] Let user specify number of OpenMP threads
Right now, OpenMP tries to use the maximum number of threads (usually equal to the number of cores). Make it so that a model takes a `n_threads` parameter, which is then passed down to `prange`.

https://stackoverflow.com/a/1006337
https://github.com/dipy/dipy/blob/master/dipy/utils/omp.pyx

WadeVaresio:
## Description

This PR adds the ability to specify the number of threads to use during parallelization.

Closes #263

## Type of Change

## Checklist

### DIFF
--- a/pulse2percept/models/_beyeler2019.pyx
+++ b/pulse2percept/models/_beyeler2019.pyx
@@ -20,7 +20,8 @@ cpdef fast_scoreboard(const float32[:, ::1] stim,
                       const float32[::1] xgrid,
                       const float32[::1] ygrid,
                       float32 rho,
-                      float32 thresh_percept):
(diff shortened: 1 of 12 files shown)


mbeyeler (changes requested):
This all looks good, but what's missing is a test case to make sure the right number of threads is being used to execute the loop. We can discuss today what such a test might look like - maybe assuming that execution time scales (roughly) linear with number of threads?

mbeyeler (approved):

PR #458, opened 2022-02-01
[ENH] Faster VideoStimulus Encode function

### LINKED ISSUE #377: [ENH] Speed up VideoStimulus
`VideoStimulus` operations are too slow and take up too much memory:

1. Don't convert everything to float arrays right away. Though the issue with that is that `rgb2gray` automatically converts to float, so then sometimes the grayscale range is [0,255] and other times it's [0,1].
2. A similar problem exists for skimage's `resize`, though there is a `preserve_range` option
3. Consider just-in-time computations and cached results
4. If Point 2 doesn't help, then cythonize

narenberg:
## Description

Attempt to speed up VideoStimulus encode function by cythonizing and optimizing loops. All tests pass, speedup is minimal. It is worth wondering whether the change is worth merging into master.

Closes #377 and #462

## Type of Change

## Checklist


pulse2percept/stimuli/_videos.pyx
```diff
@@ -0,0 +1,86 @@
...
+    if not c_isclose(np.abs(enc_data).max(), 0):
+        enc_data /= np.abs(enc_data).max()
+    out_stim = {}
+    for px_data, e in zip(vid_data, stim.electrodes):
+        px_stim = None
+        # For each pixel, we get a list of grayscale values (over time):
+        for px in px_data:
+            # Amplitude modulation:
+            amp = px * (amp_max - amp_min) + amp_min
+            s = Stimulus(amp * enc_data, time=pulse.time, electrodes=e)
```
mbeyeler, line 80: The problem with this is that you have Python overhead in the loop (from calling `Stimulus`), so I wouldn't expect this to be much faster than pure Python...

mbeyeler (commented):
Let's discuss this at our next meeting. The problem currently is that there are still a lot of necessary Python calls in the loop (e.g., `Stimulus`, `BiphasicPulse`), so you can't release the gil. This would explain why the speedup is limited... 

Making this work is tricky. You'd have to strip all the Python overhead and e.g. purely operate on arrays. However, currently the array changes shape on each iteration, which is also slow.

One way might be to:

1.  Do all the Python/error checks outside the Cython function
2.  Use p2p to create `BiphasicPulse` outside Cython. Then only pass the time & data arrays of the `BiphasicPulse` object
3.  Preallocate the time & data arrays of the final `Stimulus` that will be created
4.  Fill the values of the arrays in the loop without calling out to Python objects

narenberg:
I'm still not sure how necessary this PR is since the majority of the VideoStatus encoding test time came from creating the BostonTrain Stimulus.

pulse2percept/models/granley2021.py
```diff
@@ -582,14 +586,18 @@ def predict_percept(self, implant, t_percept=None):
...
+                for i, (ele, params) in enumerate(implant.stim.metadata
+                                                ['electrodes'].items()):
+                    if (params['type'] != BiphasicPulseTrain or
+                            params['metadata']['delay_dur'] != 0) and \
+                            np.any(implant.stim[i]):
+                        raise TypeError(
+                            f"All stimuli must be BiphasicPulseTrains with no " +
+                            f"delay dur (Failing electrode: {ele})")
+            except KeyError:
+                raise KeyError(f"All stimuli must be BiphasicPulseTrains with no " +
```
jgranley, line 599: I think this should be a type error: The problem is that the passed stimuli was not the correct type

jgranley (changes requested):
Looks good, except maybe change the error given to the user to be a typeerror instead of a key error. this is more useful to the user, otherwise they might be confused why they get a key error when they never even worked with any dictionaries

jgranley (commented):
The code all looks good to me. I am curious what the speedup is now. Like Nick mentioned, is it worth adding more cython / complication if it doesn't speed up?

Either way, I think removing the lines that check stimulus metadata (stimuli/base.py 248-252) is good. I only added this so that metadata was always collected and had a unified structure across stimuli. In this situation, the metadata is not even used anywhere, so theres no point in collecting it if its slow

mbeyeler:
The reason it's not much faster is that almost every Cython line has Python overhead. But there should be a faster way - let me see what I can do

mbeyeler:
If you use an outer product instead of all the concatenate, stim append, and dictionary business, you can go down from:

```
vid = p2p.stimuli.BostonTrain(as_gray=True, resize=(20, 20))
%timeit -n3 -r3 vid.encode()
6.54 s ± 166 ms per loop (mean ± std. dev. of 3 runs, 3 loops each)
```

to

```
vid = p2p.stimuli.BostonTrain(as_gray=True, resize=(20, 20))
%timeit -n3 -r3 vid.encode()
1.99 ms ± 405 µs per loop (mean ± std. dev. of 3 runs, 3 loops each)
```

No Cython required!
Now even the full-size RGB `BostonTrain` encodes in 1.35s, and that's on my laptop.

PR #466, opened 2022-02-24
[ENH] Deep Copyable Models

### LINKED ISSUE #392: [BUG] Corrupted stimuli when using multi-electrode Stimulus
When constructing a Stimulus object from multiple stimuli on different electrodes, sometimes the individual stimuli objects get corrupted.

For example:

```
a = BiphasicPulseTrain(20, 1, 0.45)
b = BiphasicPulseTrain(30, 1, 0.45)
b.plot(time=(25, 75))
```
image
```
stim = Stimulus({"A2":a, "A10":b})
stim.plot(time=(25, 75))
```
image

This doesn't always occur, but consistently happens with certain stimuli frequency/pulse duration combinations. Multiple time points are off by ~0.3, so it doesn't seem to be floating point error during stimulus combination or anything like that.

### LINKED ISSUE #421: [ENH] Make models copyable
Currently, it is hard to create a copy of a model object. Maybe there is already a way to do this, but I couldn't find an elegant one. The easiest current way seems to be making a new object and manually copying all of the custom parameters to the new object. 

`copy.deepcopy(AxonMapModel())` returns a `RecursionError` due to the `getattr` / `setattr` functions that distribute parameters to either the spatial or temporal model.

Potential solutions:
- Override `__copy__` and `__deepcopy__`  with custom copy function

WadeVaresio:
## Description

This PR adds the ability to perform deep copies of all models. Additionally, the equality operator is overridden for certain classes as needed in order to support checking whether a deep-copied model is truly equal to it's original

Closes #421 

## Type of Change

## Checklist


pulse2percept/models/beyeler2019.py
```diff
@@ -105,6 +106,27 @@ def _predict_spatial(self, earray, stim):
...
+
+        Returns
+        -------
+            Deep copy of the object
+        """
+        if id(self) in memodict:
+            return memodict[id(self)]
+        attributes = copy.deepcopy(self.__dict__)
+        # Remove attributes set internally
+        attributes.pop('_is_built')
```
jgranley, line 125: Why do we not set `_is_built ` and `grid` again after popping them from attributes? Will this result in a built scoreboard model not being built after copying?

pulse2percept/models/beyeler2019.py
```diff
@@ -854,6 +905,28 @@ def plot(self, use_dva=False, style='hull', annotate=True, autoscale=True,
             ann.set_yticks([])
         return ax
 
+    def __deepcopy__(self, memodict={}):
```
jgranley, line 908: It seems that this is a copy of `SpatialModel.__deepcopy__()`

pulse2percept/models/beyeler2019.py
```diff
@@ -961,3 +1034,32 @@ def predict_percept(self, implant, t_percept=None):
                                  f"built for {self.spatial.eye}.")
         return super(AxonMapModel, self).predict_percept(implant,
                                                          t_percept=t_percept)
+
+    def __deepcopy__(self, memodict={}):
```
jgranley, line 1038: It seems that this is the common copy format for all `Model` objects. Perhaps we could implement this once in `Model` and not have to reimplement it in subclasses?

pulse2percept/models/base.py
```diff
@@ -417,6 +418,53 @@ def plot(self, use_dva=False, style='hull', autoscale=True, ax=None,
             ax.set_ylabel('y (microns)')
         return ax
 
+    def __eq__(self, other):
```
jgranley, line 421: It looks like both __eq__ and __deepcopy__ are almost the same for both SpatialModel and Temporal model. Both inherit from BaseModel, so would it be possible to make changes there instead?

pulse2percept/models/base.py
```diff
@@ -417,6 +418,53 @@ def plot(self, use_dva=False, style='hull', autoscale=True, ax=None,
...
+        ----------
+        other: SpatialModel
+            SpatialModel to compare with
+
+        Returns
+        -------
+        bool:
+            True if the compared objects have identical attributes, False otherwise.
+        """
+        if not isinstance(other, SpatialModel):
```
jgranley, line 436: This looks like the only line not shared between `SpatialModel` and `TemporalModel` `__eq__`. Checking to see if other is an instance of `self.__class__` might work to unify the two functions in `BaseModel`

pulse2percept/models/granley2021.py
```diff
@@ -90,6 +91,27 @@ def __call__(self, freq, amp, pdur):
         F_bright = self.predict_freq_amp(amp * self.scale_threshold(pdur), freq)
         return F_bright
 
+    def __eq__(self, other):
```
jgranley, line 94: This also seems like one `__eq__` function in BaseModel could cover this and all of the other effect models

pulse2percept/models/beyeler2019.py
jgranley: I think this would be solved by using the parent `BaseModel` `deepcopy` which copies all attributes manually after creation

jgranley (changes requested):
Good work! It looks like much of this code could potentially be unified into a few common `__eq__` and `__deepcopy__` functions, one in `BaseModel`, and one in Model. In `BaseModel`, an empty self.__class__() object could be made and all of the attributes manually set, which I think should also  the `_is_built` flag being flipped. The `Model` deepcopy could be exactly what is currently in `ScoreboardModel`, `AxonMapModel`, etc. Do you think this would be possible? I added a few more details in the line comments

pulse2percept/models/beyeler2019.py
WadeVaresio: This is a good point. Thank you for the feedback, I will work on integrating this feedback, as well as the other feedback.

jgranley (approved):

PR #493, opened 2022-05-26
[ENH] Added NoisyWatsonMap as subclass of VisualFieldMap to simulate natural deviations in retinal coorinate to dva translation

### LINKED ISSUE #274: [ENH] Add Gaussian noise to RetinalCoordTransform
Add noise to the conversion of retinal location to visual field location (in the `RetinalCoordTransform` base class) to account for the fact that phosphenes don't always appear exactly where they're supposed to.

narenberg:
## Description

Please include a summary of the change and which issue it closes below. Include relevant motivation and context.

Closes #274 

## Type of Change

## Checklist

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/pulse2percept/utils/__init__.py
+++ b/pulse2percept/utils/__init__.py
@@ -19,6 +19,7 @@ from .base import (PrettyPrint, FreezeError, Frozen, Data, bijective26_name,
                    cached, gamma)
 from .geometry import (Grid2D, VisualFieldMap, Curcio1990Map,
                        Watson2014Map, Watson2014DisplaceMap,
+                       NoisyMap,
                        cart2pol, pol2cart, delta_angle)
 from .array import is_strictly_increasing, radial_mask, sample, unique
 from .images import center_image, scale_image, shift_image, trim_image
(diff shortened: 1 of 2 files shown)


mbeyeler (changes requested):
Let's discuss tomorrow if there's a way to add a noise option to the existing `VisualFieldMap` class or if we should wrap the latter.

But, at the very least, this needs a test case.

mbeyeler:
Hi @narenberg, I would much prefer "noise" being an attribute of any `VisualFieldMap` (see my latest commit). Still needs a test and more docstrings.

Currently, this won't work with `Watson2014DisplaceMap`, because its `dva2ret` depends on `Watson2014Map`. These two classes should just be fused - but I can do that later. I'm mainly putting this here to remind my future self of it.

PR #495, opened 2022-06-22
[ENH] Allow easy stimulus coloring in implant.plot()

jgranley:
## Description
This introduces an easy way to color the electrode array in `implant.plot()` based on the amplitude of the stimulating electrodes. 
`implant.plot()` now accepts a `color_stim` parameter, defaulting to False. If True, the amplitude of each electrode is mapped to the color of each active electrode in the plot. The color map can also be specified with the `cmap` parameter.

I'm currently not sure what the best default colormap to use is. 
**Example**: 
Default behavior unchanged
```
implant = ArgusII()
implant.stim = {'B4': BiphasicPulseTrain(5, 2, 1), 'C5': BiphasicPulseTrain(5, 3, 1), 'D6': BiphasicPulseTrain(5, 4, 1)} 
implant.plot()
```
image

A single monotonically increasing cmap (e.g. 'OrRd') could make sense:
```
implant.plot(color_stim=True, cmap='OrRd')
plt.colorbar()
```
image

Or having different colors for each unique amp also makes sense (e.g. 'gist_rainbow')
image

Note: The colorbar is not added by default, only if the user calls `plt.colorbar`

## Type of Change

## Checklist

### DIFF
--- a/pulse2percept/implants/base.py
+++ b/pulse2percept/implants/base.py
@@ -170,7 +170,7 @@ class ProsthesisSystem(PrettyPrint):
             raise ValueError(err_str)
         return stim
 
-    def plot(self, annotate=False, autoscale=True, ax=None, color_stim=False, cmap='OrRd'):
+    def plot(self, annotate=False, autoscale=True, ax=None, stim_cmap=False):
         """Plot
 
         Parameters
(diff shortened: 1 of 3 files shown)


mbeyeler (changes requested):
Looks great! But can we combine the two new keyword arguments for simplicity? e.g. `stim_cmap`:
- if it's False, don't use color
- if it's True, use standard colormap (`hot` or `viridis` would be good)
- if it's a string or object, assume it's a Matplotlib colormap

jgranley:
Updated to only have the one parameter in `implant.plot()`. Defaults to hot colormap. `earray.plot `still takes two parameters, because it needs both the actual stimulus object as well as the colormap.

jgranley (commented):
Updated to only have the one parameter in `implant.plot()`. Defaults to hot colormap. `earray.plot` still takes two parameters, because it needs both the actual stimulus object as well as the colormap.

PR #497, opened 2022-06-26
[ENH] Only run predict percept on unique time points

### LINKED ISSUE #456: [ENH] Only evaluate models on unique amplitudes
Currently, all models (except `BiphasicAxonMap`) by default run `predict_spatial` on every time point. For standard pulse trains, this repeats a lot of computation. Instead, models should only run `predict_spatial` on unique amplitudes.  

**Describe the solution you'd like**
Change [`SpatialModel.predict_percept` ](https://github.com/pulse2percept/pulse2percept/blob/master/pulse2percept/models/base.py#L264) to find unique amplitudes / corresponding time points, change the stim object passed to `_predict_spatial` to only have these amplitudes, and then map the calculated response back to all of the corresponding original times.

Note that in the case of multiple electrodes, the set of amplitudes on each electrode must be unique in order for the it to be considered a unique timepoint. This check for uniqueness will have to be fast.

jgranley:
## Description
This changes `SpatialModel.predict_percept` to only call `_predict_spatial` for a compressed stimulus where the distribution of amplitudes across electrodes is unique. 

Performance improvement with default `AxonMapModel` on my laptop:

- With 20Hz `BiphasicPulseTrain`: 6.5ms vs 309ms, **~50x Speedup**
- With 250Hz `BiphasicPulseTrain`: 57ms vs 5s, **~90x Speedup**
- With random frequency `BiphasicPulseTrain` on 60 electrodes: 13.5min vs 4min , **~3.5xSpeedup**

Closes #456 

## Type of Change

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pulse2percept/models/base.py
+++ b/pulse2percept/models/base.py
@@ -346,11 +346,19 @@ class SpatialModel(BaseModel, metaclass=ABCMeta):
             # Calculate the Stimulus at requested time points:
             if t_percept is not None:
                 # Save electrode parameters
-                stim = Stimulus(stim)  # make sure stimulus is in proper format
+                if not isinstance(stim, Stimulus):
+                    stim = Stimulus(stim)  # make sure stimulus is in proper format
                 stim = Stimulus(stim[:, t_percept].reshape((-1, n_time)),
                                 electrodes=stim.electrodes, time=t_percept,
                                 metadata=stim.metadata)
-            resp = self._predict_spatial(implant.earray, stim)
+            # find unique stimulus points
+            amps, t_unique, inverse = np.unique(stim.data.T, axis=0, 
+                                            return_index=True, return_inverse=True)
+            stim_unique = Stimulus(stim[:, t_unique], electrodes=stim.electrodes,
+                                   time=t_unique)
+            resp_unique = self._predict_spatial(implant.earray, stim_unique)
+            resp = resp_unique[..., inverse]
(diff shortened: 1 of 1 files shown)


pulse2percept/models/base.py
```diff
@@ -346,11 +346,21 @@ def predict_percept(self, implant, t_percept=None):
             # Calculate the Stimulus at requested time points:
             if t_percept is not None:
                 # Save electrode parameters
-                stim = Stimulus(stim)  # make sure stimulus is in proper format
+                if not isinstance(stim, Stimulus):
+                    stim = Stimulus(stim)  # make sure stimulus is in proper format
```
mbeyeler, line 350: I know `stim = Stimulus(stim)` was here before, but is it actually needed? It looks like `stim` is created from a deepcopy of `implant.stim` and there's calls to e.g. `stim.is_compressed` before, so I think we can safely remove these two lines here

mbeyeler (commented):
Fantastic! One small comment that's not even your fault

pulse2percept/models/base.py
jgranley: Good catch. This was added in #396 to make sure the stimulus metadata was associated with each electrode for the `BiphasicAxonMapModel`. But since then, `BiphasicAxonMapModel ` has been changed to have its own `spatial.predict_percept`, so this is obsolete now

mbeyeler (approved):

PR #509, opened 2023-01-17
[ENH][REF] Modified Grid2D and VisualFieldMap to support multiple visual areas

jgranley:
## Description

This includes the base functionality needed for allowing cortical models:

1. `Grid2D` (`model.grid`) now supports multiple visual areas. This is sort of the "Anatomy" class now. There are three major changes

    - Internally, all layers are now stored in two dicts, `grid._x` and `grid._y`. These contain the coordinate grids for each visual area, or _region_.  For example, `model.grid._x['ret']` stores the x coordinates in the retinal region. 
    - Each layer can still be accessed in the same way as before, (e.g. `model.grid.xret`). A set of [helper properties](https://github.com/pulse2percept/pulse2percept/pull/509/files#diff-6491c3d72a4ecc157d02d57fd2cad9233196d5df010e84d8a470e765dedc16e1R62-R103) allows for any registered region to be accessed through this convenient syntax. This is backwards compatible with retinal models, which call `model.grid.xret`, and convenient for future cortical models, which can call `model.grid.xv1`, etc. A `ValueError` is raised if the current grid doesn't have the specified region (such as if a retinal model calls `grid.xv1`).
    - The model's `VisualFieldMap`, or retinotopy, determines for which regions coordinate grids will be instantiated. `model.build()` now calls `grid.build(retinotopy)`, which computes the mapping(s) between dva and the regions specified by the retinotopy, and stores the result in the grid.

2. `VisualFieldMap` inheritance structure changed to support multiple visual areas

    - `Grid2D` now requires that each retinotopy provides `region_mappings` and `inv_region_mappings` functions, which return a dict of region name -> transform_function. 
    - A new base class for retinal retinotopies, `RetinalMap`, simply provides the existing `ret2dva`/`dva2ret` functions. Retinal retinotopies should inherit from this
    - A new base class for cortical retinotopies, `CorticalMap`, allows for a list of visual regions to be specified, and maps them per layer. Subclasses should override `dva2v1` and associated functions. `V1V2V3Map` would perhaps be a more accurate name.

With these changes, any amount of new regions are possible. Individual models are still responsible for filling in whatever details they want (e.g. axon map, orientation columns), but there is now a flexible system for handling new and multiple visual areas.

## Restructuring
I think the next thing to consider is how to restructure the repo to accommodate the new cortical classes.  I think `utils/geometry` should by upgraded, since there are about to be quite a few retinotopies and Grid is a pretty central part of models. Potentially a submodule under models (e.g. `models/topography`?)

In general though, there is going to be a lot more cortical stuff coming in. There are three options:

~~1. One central `cortex` module. This is convenient and clear to users. However, it doesn't make the most sense, since a lot of the cortical stuff will be the same as existing (e.g. we have models, implants, retinotopies, etc). We might have to have submodules for each of these again (e.g. `cortex.models`, `cortex.implants`, etc)~~

**2. Submodules within each existing module (e.g. `models.cortex`).** 

~~3. No explicit cortex module, just throw them all in in their own `.py ` files in existing modules.~~

## Type of Change

## Checklist

Added tests for new grid capabilities. 

## Changes

- [x] grid.v1.x (Named tuple)!
- [x] maps_to_dva

Future:
- topography module
- submodule for models.cortex and implants.cortex


pulse2percept/utils/geometry.py
```diff
@@ -221,24 +283,110 @@ def plot(self, transform=None, label=None, style='hull', autoscale=True,
...
+
+    @abstractmethod
+    def v22dva(self, x, y):
+        """Convert V2 coordinates (um) to degrees visual angle (dva)"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def v32dva(self, x, y):
+        """Convert V3 coordinates (um) to degrees visual angle (dva)"""
+        raise NotImplementedError
```
jgranley, line 387: This naming is a bit unfortunate

pulse2percept/utils/geometry.py
```diff
@@ -221,24 +283,110 @@ def plot(self, transform=None, label=None, style='hull', autoscale=True,
...
+    @classmethod
+    def region_mappings(cls):
+        return {'ret' : cls.dva2ret}
+    
+    @classmethod
+    def inv_region_mappings(cls):
+        return {'ret' : cls.ret2dva}
+    
+    @staticmethod
     @abstractmethod
```
jgranley, line 314: Not completely sure why all the retinal retinotopy transforms are static. Models always instantiate them anyways. Cortical transforms require state so this pattern doesn't work there

pulse2percept/utils/geometry.py
```diff
@@ -221,24 +284,106 @@ def plot(self, transform=None, label=None, style='hull', autoscale=True,
 
 
 class VisualFieldMap(object, metaclass=ABCMeta):
-    """Base class for a visual field map (retinotopy)
+    """ Base template class for a visual field map (retinotopy) """
 
-    A template
+    @abstractmethod
+    def region_mappings(self):
```
mbeyeler, line 290: `region_mappings` and `inv_region_mappings` is a bit confusing to a novice, because the direction of the mapping is not clear. Maybe make the name more explicit, e.g. `maps_from_dva`

pulse2percept/utils/geometry.py
```diff
@@ -221,24 +284,106 @@ def plot(self, transform=None, label=None, style='hull', autoscale=True,
...
+        if 'v3' in self.regions:
+            mappings['v3'] = self.dva2v3
+        return mappings
+    
+    def inv_region_mappings(self):
+        mappings = dict()
+        if 'v1' in self.regions:
+            mappings['v1'] = self.v12dva
+        if 'v2' in self.regions:
+            mappings['v2'] = self.v22dva
```
mbeyeler, line 354: :)

jgranley:
This PR is getting monolithic, so I'll do restructuring in a different PR

PR #523, opened 2023-02-09
[FIX] Fix electrode numbering annotation in implant.plot()

jgranley:
## Description
Right now electrode numbers appear under the electrode patches, so you cant really ever see them unless the plot is huge enough that the full name fits inside the electrode. If a user is taking the time to specify `annotate=True`, then annotations should always be legible.

**Old**:
image

**New:**
image

## Type of Change

### DIFF
--- a/pulse2percept/implants/electrode_arrays.py
+++ b/pulse2percept/implants/electrode_arrays.py
@@ -172,16 +172,18 @@ class ElectrodeArray(PrettyPrint):
                 # Regular use case: single object
                 patches.append(electrode.plot_patch((electrode.x, electrode.y),
                                                     **kwargs))
-            if annotate:
+            
+        patch_collection = PatchCollection(patches, match_original=True,
+                                          zorder=ZORDER['annotate'], cmap=cm, norm=norm)
+        ax.add_collection(patch_collection)
+        ax._sci(patch_collection) # enables plt.colormap()
+        if annotate:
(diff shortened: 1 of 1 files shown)


pulse2percept/implants/electrode_arrays.py
```diff
@@ -179,7 +179,7 @@ def plot(self, annotate=False, autoscale=True, ax=None, color_stim=None, cmap='O
                               'fc': (1, 1, 1, 0.7)},
                         zorder=ZORDER['annotate'])
         patch_collection = PatchCollection(patches, match_original=True,
-                                          zorder=ZORDER['annotate'], cmap=cm, norm=norm)
+                                          zorder=ZORDER['foreground'], cmap=cm, norm=norm)
         ax.add_collection(patch_collection)
```
mbeyeler, line 183: I don't get why this stopped working. "annotate" has zorder=2000, and "foreground" has zorder=50. There's also "front" that could be used with zorder=9999

jgranley: I think the problem was that both the patches and the annotations were registered with the annotate zorder before, so the patches went on top because that happens later. 

I changed it so that the patches are in the foreground now instead of annotate layer

mbeyeler: Oh, gotcha. That makes sense then

mbeyeler (approved):

PR #531, opened 2023-03-07
[ENH] Add Polimeni visual field mapping

### LINKED ISSUE #511: [ENH] Add Polimeni visual field mappings
## Description
The [Polimeni](https://www.sciencedirect.com/science/article/pii/S0042698906001428) visual field maps describe the retinotopic transform between points in visual field (degrees visual angle) and location in V1/V2/V3 (um). This can be added to the `utils/geometry.py` file. 

There is an existing implementation in [p2pcortex](https://github.com/bionicvisionlab/p2pcortex/blob/master/p2pcortex/retinotopy.py). However, a number of issues need to be resolved before it can be added to p2p:

### The model must be refactored to subclass [CorticalMap](https://github.com/pulse2percept/pulse2percept/blob/cortex/pulse2percept/utils/geometry.py#L334).
This means it will need to implement `dva_to_v1 `, `v1_to_dva`, etc. instead of the existing `dva2v1` etc. It also needs to remove `ret2dva` `dva2ret` functions, which is no longer needed. 

### Inverse mapping for V3 (`v3_to_dva`)
There are currently inverse mappings for V1 and V2, but not V3. There is a `dva2v3` though, so to invert it you basically have to do everything in reverse, similar to the implementations for V1 and V2.

### The `eye` parameter should be removed. 
The Polimeni mapping only works for the right hemisphere. Currently the way that this works is if the 'eye' attribute (an old name denoting which hemisphere of visual cortex) is 'RE', then the retinotopy works as normal. If `eye` is 'LE', then the points to be mapped are inverted across the y axis, the same mapping functions are applied, and then the result is flipped back across the y axis.

This current implementation only allows for the retinotopy to work for one of the two hemispheres at once. A more flexible method would be to detect which points are in the left vs right hemispheres (right hemisphere = left visual field = negative y coordinate in DVA).  Then use the normal transform for all right hemisphere points, and do the same invert y -> transform -> invert y procedure for the left hemisphere.

Basically, we just want to make the retinotopy work for all points, regardless of whether they are mapping to the right or left hemisphere.

With the current implementation, this _almost_ works already. The `_correct_eye_*` methods do the inversion. These could simply be refactored to only invert the points in the left hemisphere (instead of relying on the `eye` attribute). These should also be renamed (e.g. to something like `_invert_left`, anything without 'eye' in the name).

### Double check points are mapping to correct visual field
Also, in the current implementation, I think that the mappings aren't corresponding to the correct visual field (e.g. it assumes points in the left visual field map to the left hemisphere, when they should map to the right). We should check and fix this if necessary

### Test cases
There should be some test cases ensuring (1) that the mapping gives the correct result and (2) that the inverse mappings (`*_to_dva`) are actually the inverse of the forward (dva_to_*) mappings

There are a couple of other issues related to this such as [plotting updates](https://github.com/bionicvisionlab/p2pcortex/issues/12), but we will fix those in separate issue. Here, focus on core functionality.

## Relevant links:

- [p2pcortex implementation](https://github.com/bionicvisionlab/p2pcortex/blob/master/p2pcortex/retinotopy.py)
- [p2pcortex pull request](https://github.com/bionicvisionlab/p2pcortex/pull/15)
- [Artifacts](https://github.com/bionicvisionlab/p2pcortex/issues/9) (Not sure if this is still happening or not)

tallyhawley:
## Description

Adds the Polimeni visual field mapping to `topography/cortex.py`, as defined in [Polimeni et al. (2006)](https://pubmed.ncbi.nlm.nih.gov/16831455/). This is a retinotopic transform between points in the visual field (in degrees-visual-angle) and location in V1/V2/V3 (in um). The new `Polimeni2006Map` subclasses `CorticalMap`. Points in the left visual field are mapped to the right hemisphere in V1/V2/V3 and vice-versa, and points are inverted vertically in the visual cortex.

The right hemisphere in V1/V2/V3 is represented by positive x-values and the left hemisphere is represented by negative x-values.

Closes #511 

## Type of Change

## Checklist


jgranley:
Looks great, very thorough tests! If there are any issues I think I'll find them when I merge the Cortex Model code, which uses the retinotopy.

jgranley (commented):
Apparently I need another comment to close the review...

Issue #512, opened 2023-01-24
[ENH] Update grid.plot to take multiple transforms


jgranley:
Currently, it is hard to plot multiple visual regions, see [p2pcortex](https://github.com/bionicvisionlab/p2pcortex/blob/master/p2pcortex/models.py#L237). 

A simple solution would be to change `Grid2D.plot` in utils/geometry.py so that the `transform` parameter is allowed to be a list or dict of transforms. If it is, each of the transforms is applied and the results plotted in a different color, and labeled according to the transform name (either dict key, or str of fn name)

Output should be something like this for `style=scatter`
![image](https://user-images.githubusercontent.com/25232557/214423408-74b62d05-4f06-4f65-9f1e-7be68623c593.png)

but the other options (hull and cell) should be similar and still work for multiple regions.

See also: https://github.com/bionicvisionlab/p2pcortex/issues/14. This might be an issue with Grid2D's plotting function, not sure

Notes:
https://github.com/pulse2percept/pulse2percept/blob/cortex/pulse2percept/topography/base.py

- in constructor, set self.retinotopy = None (for later)
- in build, set self.retinotopy = retinotopy
- in plot, remove transform and label arguments
- in all cases where the function is called, remove transform and label arguments
- in plot, loop over the "if style.lower() == "cell"" if statement with the transform functions(values of self.retinotopy) and labels (keys of self.retinotopy), using matplotlib color palate to get multiple colors if needed

jgranley:
So I had to do some of this in order to test the new Scoreboard cortex model. 
After merging that code into this, self.retinotopy is already set in constructor/build.The rest of the changes needed for this issue are mostly the same, just the if statement you're going to be looping over is a bit different now. 

After this issue is resolved, we should be able to delete the function `models.cortex.ScoreboardModel.plot` and the code should still look like the following: 

**style='cell'**
![Image](https://user-images.githubusercontent.com/25232557/223626110-34016221-9fde-4177-8aac-34bfeae5f38c.png)


**style='scatter'**
![Image](https://user-images.githubusercontent.com/25232557/223625981-514cc830-527f-4052-8fd9-0f99f02d6a47.png)

**style='hull'**
![Image](https://user-images.githubusercontent.com/25232557/223626036-9f7796ab-b52d-46e1-9b56-a3a1ef7f6d26.png)


I had to add multiple (currently hacky) checks to make sure plotting looks good even with discontinuous regions. Happy to replace this with something better if we can come up with it.

PR #533, opened 2023-03-08
[ENH] Add Cortex ScoreboardModel, updates to PolimeniMap

### LINKED ISSUE #514: [ENH] Add ScoreboardCortex
Add ScoreboardCortex, a scoreboard model implementation with support for multiple visual regions.

This should be refactored from https://github.com/bionicvisionlab/p2pcortex/blob/master/p2pcortex/models.py#L15

Will require the Polimeni retinotopy to be completed.

jgranley:
## Description
Adds the scoreboard model for cortex, as well as a number of other changes to Visual Field Maps, and `BaseModel`. This PR has some somewhat hacky solutions, so some items might need further discussion/revision

Changes:

1. Add `p2p.models.cortex.ScoreboardModel  `

    It works by calculating a scoreboard current spread for each simulated region (e.g. v1), and summing the results. This is basically an abstraction of the current scoreboard model class (you can even hack it to be the same by setting `model.regions = ['ret']`). It calls the cython code for the existing scoreboard model.

    Right now the naming is the same as the original, but it is located in `models.cortex` instead of `models`. If theres a better name we could change this. It also might be worth discussing if its even worth having this as a separate model, as opposed to just making the normal scoreboard model more general. I tend to think that it is worth it though, just because having a concrete separate cortical scoreboard model seems good.

2. Various changes to `Polimeni2006Map`
    - Add `jitter_boundary` parameter, defaulting to true. The polimeni map is undefined at dva x = 0, and, for v2 and v3, when y=0. Normal `xrange/yrange/xystep` parameters will almost always have values at these points. However, they are all currently set to NAN and not simulated, which is a waste, and causes percepts to look weird. You can force odd grid spacing (e.g. `xrange=(-14.999, 15.001)`, but the the output percepts are not very clean looking because the grid ticks are in weird places. Thus, it seemed more elegant to just add a tiny constant to the grid values that are on the boundary.
    - Add NAN's where theta >**=**  $\pi$/2, or theta <**=** - $\pi$/2 (include equality)
    - Make `dva_to_*` take in cartesian coordinates, and `*_to_dva `output cartesian coordinates
    - Add checks enforcing input to be np array
    - Removed nans for points across x=0 (see change 5)

3. Make `VisualFieldMap` inherit from `BaseModel`
    This is very useful for getting frozen, pretty print, and default_params functionality (doesn't use build). However it introduces a nasty circular import error. With this change, the module dependencies are: `topography` (`base.VisualFieldMap`)  ----(depends on)--->  `models` (`base.BaseModel`) --> `implants`, `percepts`, and `topography` (`ProsthesisSystem`, `Percept`, `Grid2D`); and additionally `implants` and `percepts` modules both depend on `topography` (`Grid2D`). 

    There are a few options to resolve this, and am definitely open to discussion on which is best:
    - **a) Import dependencies at runtime (current solution).** `VisualFieldMap` cannot do this because it inherits from `BaseModel`. Instead, `models.base` now imports from `topography` at runtime. In addition, `implants` and `percepts` now also import `Grid2D` at runtime. This is somewhat inelegant, but it works. Notably, this will not affect the user, who can import anything in any order they want.
    - **b) Move `BaseModel` into utils.** This could also make sense, because multiple things (spatial/temporal models, biphasic effects models, and Visual Field Maps) all use `BaseModel`. I originally implemented this, but it just seemed wrong to have the base model class not be in `models.base`.
    - **c) Move `Grid2D` into `models.base`.** This also could make some sense, since pretty much the only things that use `Grid2D` are models. But conceptually `Grid2D` is very much a topographical objects, so staying in the topography module seems better

4. Change Grid2d.plot to look nice with discontinuous visual areas.
    This is super hacky right now, but basically it checks the name of the transform to see if one of the discontinuous regions is in the name. If so, it alters plotting behavior so the boundaries don't mess up cells/hulls. Shouldnt be hacky anymore after #512 

5. Modified how visual field maps handle both hemispheres
    TLDR: The coordinate system specified by CorticalMap has the right hemisphere fovea at (0,0), and left hemisphere fovea at (left_offset, 0) to allow simulating both hemispheres on the same cortical sheet. The model enforces that no current can flow across the hemispheres.

    Before, there was a check in `add_nans` in Polimeni that would remove any points with cortical x coordinate < 0. This was based on the assumption that the fovea was at x=0 and that you couldn't go left of the fovea in the right hemisphere. This allowed us to put both hemispheres on separate halves of the same coordinate plane.

    However.... I have since learned that the wedge dipole mapping DOES cross the boundary, and you are supposed to used separate maps for each hemisphere. Thus, we were discarding many valid points in v3 and v2.

    Implementing a separate coordinate grid for each hemisphere is probably the best solutions... but this is cumbersome, and does at all fit into the existing p2p model flow. Instead, I kept it so we do both hemispheres on one set of coordinates.  Now, `CorticalMap` is modified so that the left hemisphere fovea is always located at x = -`left_offset` (set to 20000 by default). I also modified the fast scoreboard cython so that it explicitly disallows current to spread from the right to the left hemisphere. 

    Example showing the fovea of left hemisphere is no longer at (0, 0):
image

7. A few random doc updates

Closes #514 

## Type of Change

- [x] Breaking change


jgranley:
Basically finished, but some tests are failing only on mac?? Hard to reproduce

PR #534, opened 2023-03-14
[ENH] Update grid.plot to take multiple transforms

### LINKED ISSUE #512: [ENH] Update grid.plot to take multiple transforms

lukeyoffe:
## Description

When plotting cortical models, we often want to plot multiple regions (v1, v2, and v3), but this was not possible with the existing grid.plot function
- Changed grid.plot to loop through all of the transforms in self.retinotopy rather than taking in a transform as an argument
- Added a `use_dva` flag that defaults to `False`, but when set to `True`, will not apply any transforms to preserve the ability to plot in dva coordinates (rather than retinal/cortical coordinates)

**Notebook**
It's hard to write unit tests for plotting, so here is a notebook showing that plots look correct.
[CortexPlotTester.pdf](https://github.com/pulse2percept/pulse2percept/files/11023999/CortexPlotTester.pdf)

Before / after notebooks for some other plots
[Before](https://github.com/pulse2percept/pulse2percept/files/10973411/random_tests_before.pdf)
[After](https://github.com/pulse2percept/pulse2percept/files/10973412/random_tests_updated.pdf)

Closes #512 

## Type of Change


jgranley (changes requested):
Note of changes we talked about 3/14:

- The default color of gray shouldn't change for unbuilt retinotopies, or for retinal models. I'm thinking it might just be easier to have a custom dict mapping regions to colors (i.e. 'ret' -> gray, 'v1' -> red, etc) and just grab a random color from one of the sequential colormaps if the region is unknown.
- Replace __name__ with transform

jgranley (approved):

PR #547, opened 2023-05-28
[ENH] Add Grinten / Stevenick / Lozano model (Dynaphos)

### LINKED ISSUE #526: [ENH] Add Grinten / Stevenick / Lozano Model
Add the new cortical model from the Grinten / Stevenick / Lozano paper. This will involve both a new temporal and spatial model. It might be a bit of a challenge to fit it into our existing framework, but this is one of the major additions the cortex release needs. 
Their code uses pytorch. Ours is fine to use a pytorch class too, but it should be wrapped in a `SpatialModel` / `TemporalModel` class just like all the other models.

It seems like just doing a raw python implementation first will be best. If this is way too slow, then perhaps we should use torch instead for the temporal component, or cython if we can't do anything else.

We will also have to come up with a better name, because listing all 3 first authors is super long. Maybe they call it something in the paper?

We should also add an example proving that this model achieves all the same fits to empirical data as the original paper. 

# Links

- [Twitter Thread](https://twitter.com/AntonioLozanoDL/status/1607358465550213120)
- [Paper](https://www.biorxiv.org/content/10.1101/2022.12.23.521749v1)
- [Code](https://www.biorxiv.org/content/10.1101/2022.12.23.521749v1)

tallyhawley:
## Description

Adds the Dynaphos model outlined in [Grinten, Stevenick, Lozano (2022)](https://www.biorxiv.org/content/10.1101/2022.12.23.521749v1.full.pdf).

This is a model which cannot separate its temporal and spatial components. It extends `BaseModel`. 

This is because each phosphene's brightness is not determined by instantaneous current, but by the effective current, which is affected by the memory trace (which builds up over time). This means that in order to determine the brightness at a given time step, we need to have calculated all prior time steps-- and time steps with identical instantaneous stimulus do not necessarily have identical percepts.

All model constants can be set in its constructor, but the default values are those specified in the paper.

It has a default frame rate of 50fps (dt = 20ms), but this can be set. If `t_percept` is not passed to `predict_percept`, a percept will be returned at a frame rate determined by `dt`.

The amplitude for each electrode at each time-step is determined by the maximum amplitude of the stimulus on that electrode over the duration of time step.

It takes any stimulus (with a time component), and assumes a default frequency & pulse duration of 300Hz and 0.17ms respectively. However, if the stimulus is a `BiphasicPulseTrain`, it will use the frequency and pulse duration specified in the stimulus metadata.

Activation and brightness trends are very similar to those seen in the referenced paper:
image
_Pulse2percept implementation of the Dynaphos model_
image
_Original implementation of the Dynaphos model_

Timing information (time to run `predict_percept` on a 1sec stimulus).
Model is built with default parameters.
Grid side length was set using `xystep`, and FPS was set using `dt`.
image

Closes #526 

## Type of Change

## Checklist

- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/pulse2percept/models/cortex/dynaphos.py
+++ b/pulse2percept/models/cortex/dynaphos.py
@@ -1,117 +1,118 @@
 """`DynaphosModel`, `DynaphosSpatial`, `DynaphosTemporal`"""
 import numpy as np
 import warnings
-from ..base import Model, SpatialModel
+from ..base import Model, SpatialModel, TemporalModel
 from ...utils import cart2pol
 from ...topography import Polimeni2006Map
 
 class DynaphosSpatial(SpatialModel):
-   """Adaptation of the Dynaphos model from Grinten, Stevenick, Lozano (2022)
+    """Adaptation of the Dynaphos model from Grinten, Stevenick, Lozano (2022)
 
-   Implements the spatial component of the Dynaphos model. Percepts from each
-   electrode are Gaussian blobs, with the size dictated by a magnification factor
-   M determined by the electrode's position in the visual cortex.
-   """
+    Implements the spatial component of the Dynaphos model. Percepts from each
+    electrode are Gaussian blobs, with the size dictated by a magnification factor
+    M determined by the electrode's position in the visual cortex.
+    """
 
-   def __init__(self, **params):
-         self._regions = None
-         super(DynaphosSpatial, self).__init__(**params)
+    def __init__(self, **params):
+            self._regions = None
+            super(DynaphosSpatial, self).__init__(**params)
 
-         # Use [Polemeni2006]_ visual field map with parameters specified in the paper
-         self.retinotopy = Polimeni2006Map(a=0.75,k=17.3,b=120,alpha1=0.95,regions=self.regions)
-   
-   def get_default_params(self):
-        """Returns all settable parameters of the scoreboard model"""
-        base_params = super(DynaphosSpatial, self).get_default_params()
-        params = {
-                    'xrange' : (-5, 5),
-                    'yrange' : (-5, 5),
-                    'xystep' : 0.1,
-                    # Visual field regions to simulate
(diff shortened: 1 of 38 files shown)


pulse2percept/models/cortex/dynaphos.py
```diff
@@ -0,0 +1,327 @@
...
+        if not isinstance(regions, list):
+            regions = [regions]
+        self._regions = regions
+
+    def __init__(self, **params):
+            self._regions = None
+            super().__init__(**params)
+
+            # Use [Polemeni2006]_ visual field map with parameters specified in the paper
+            self.retinotopy = Polimeni2006Map(a=0.75,k=17.3,b=120,alpha1=0.95,regions=self.regions)
```
jgranley, line 37: Move to default params

pulse2percept/models/cortex/tests/test_dynaphos.py
```diff
@@ -0,0 +1,92 @@
...
+    for amp in np.linspace(20, 70, 5):
+        # implant.stim = PulseTrain(model.dt, freq=20, amp=amp, dur=sdur,
+        #                           pulse_dur=pdur, interphase_dur=pdur)
+        implant.stim = BiphasicPulseTrain(20, amp, pdur, interphase_dur=pdur,
+                                          stim_dur=sdur)
+        percept = model.predict_percept(implant, t_percept=t_percept)
+        bright_amp.append(percept.data.max())
+    bright_amp_ref = np.array([0.0, 0.165, 0.262, 0.391, 0.536])
+    npt.assert_almost_equal(bright_amp, bright_amp_ref, decimal=3)
+
```
jgranley, line 74: Add test from granley2021 with number of points greater than threshold

jgranley:
Looks great! Looks likes its really fast too, was not expecting that with the raw python, but must be a good implementation

PR #556, opened 2023-06-13
[FIX] scikit-image/imageio bugfixes (kwargs, saving float32 to png, unexpected saved percepts)

### LINKED ISSUE #505: [BUG] Saving percept gives unexpected results
**Describe the bug**
`percept.save()` results in videos that look different from `percept.plot()` and `percept.play()`. Not completely sure of the cause yet, but I suspect it has something to do with normalization (e.g. `vmin`, `vmax`) or `uint8` conversion 

**To Reproduce**
```
import pulse2percept as p2p 
m = p2p.models.Model(spatial=p2p.models.AxonMapSpatial(), temporal=p2p.models.FadingTemporal())
m.build()
stim = p2p.stimuli.Stimulus({'A3' : p2p.stimuli.BiphasicPulseTrain(20, 1, 0.45)})
p = m.predict_percept(p2p.implants.ArgusII(stim=stim))
```

`percept.plot()`:
image

```
from IPython.display import Video
p.save('test.mp4')
Video("test.mp4")
```
image

**System specifications (please complete the following information):**
Python 3.10 Ubuntu 18.04

### LINKED ISSUE #549: [BUG] ImageStimulus constructor passes unexpected arg to imread()
**Describe the bug**
The ImageStimulus constructor, when initializing from a filename, calls `imread(source, format=format)`. `format` is not a recognized keyword for `imread` nor the plugin that it calls. This is causing 17 ImageStimulus unit tests to fail, both locally and in CI/CD.

**To Reproduce**
```python
from pulse2percept.stimuli import LogoBVL
stim = LogoBVL()
```

**Expected behavior**

**Screenshots**
image

**System specifications (please complete the following information):**
- OS: macOS Monterey 12.4
- Python version: 3.10.11
- pulse2percept version: 0.9.0.dev0

tallyhawley:
## Description

This PR fixes several issues.

1. When loading an `ImageStimulus`, we were passing a `format` keyword. This is not supported in newer versions of skimage & imageio, and has been removed. (Originally replaced by the `extension` keyword but this is not supported by the TIFF image loader for some reason). (Bug referenced in #549) 
2. We were unable to save ImageStimulus (which has dtype `float32`) as PNG, JPEG, etc. For these file-types, we scale to `uint8` before saving and output a warning that the data has been scaled/compressed. TIFF files are saved with full original precision.
3. When saving a video percept, we were passing a `fps` keyword. This is no longer supported by imageio for saving GIFs & other miscellaneous file types. This has been supplemented by a `duration` keyword argument (set to 1000 / fps), which is passed if we fail when passing the `fps` keyword, as `fps` is still used by FFMPEG when saving MP4, WAV, etc.
4. When saving a percept, the saved image or video was corrupted (as described in #505). This has been fixed by using skimage to scale percepts directly to the UINT8 range instead of scaling to UINT16 then casting to UINT8.
Recreating the same percept as in the issue we have:
image

Closes #505 
Closes #549

## Type of Change

## Checklist

### DIFF
--- a/pulse2percept/percepts/base.py
+++ b/pulse2percept/percepts/base.py
@@ -378,9 +378,9 @@ class Percept(Data):
             imageio.imwrite(fname, img_as_ubyte(data).squeeze(2))
         else:
             # Throw error if we try to save as a static image
-            # if any(fname.endswith(ext) for ext in 
-            #        ['.jpg','.jpeg','.bmp','.png','.tif','.tiff','.jif','.jfif']):
-            #     raise ValueError(f"Cannot save multi-frame percept as a static image: {fname}")
+            for ext in ['.jpg','.jpeg','.bmp','.png','.tif','.tiff','.jif','.jfif']:
+                if fname.endswith(ext):
+                    raise ValueError(f"Cannot save multi-frame percept as a static image: {fname}")
             # With time component, store as a movie:
             if fps is None:
                 interval = unique(np.diff(self.time))
--- a/pulse2percept/percepts/base.py
+++ b/pulse2percept/percepts/base.py
@@ -378,9 +378,9 @@ class Percept(Data):
             imageio.imwrite(fname, img_as_ubyte(data).squeeze(2))
         else:
             # Throw error if we try to save as a static image
-            if any(fname.endswith(ext) for ext in 
-                   ['.jpg','.jpeg','.bmp','.png','.tif','.tiff','.jif','.jfif']):
-                raise ValueError(f"Cannot save multi-frame percept as a static image: {fname}")
+            # if any(fname.endswith(ext) for ext in 
+            #        ['.jpg','.jpeg','.bmp','.png','.tif','.tiff','.jif','.jfif']):
+            #     raise ValueError(f"Cannot save multi-frame percept as a static image: {fname}")
             # With time component, store as a movie:
             if fps is None:
                 interval = unique(np.diff(self.time))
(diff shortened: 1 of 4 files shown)


pulse2percept/percepts/base.py
```diff
@@ -396,6 +400,9 @@ def save(self, fname, shape=None, fps=None):
                     if h % VIDEO_BLOCK_SIZE > 0:
                         out_h += VIDEO_BLOCK_SIZE - (h % VIDEO_BLOCK_SIZE)
                     data = resize(data, (out_h, out_w))
-            data = img_as_uint(data).astype(np.uint8)
-            imageio.mimwrite(fname, data.transpose((2, 0, 1)), fps=fps)
+            data = img_as_ubyte(data)
+            try:
+                imageio.mimwrite(fname, data.transpose((2, 0, 1)), fps=fps)
+            except:
```
jgranley, line 406: Do we know what error is thrown here? Would be better to either only catch the right error, or ideally, do the switch beforehand based on when its necessary (guessing its determined by filetype).

jgranley:
Looks great, thanks for the hotfix! Only thing I think is maybe we can make the exception catching more specific, so we aren't doing that if there's a e.g. permission error.

pulse2percept/percepts/base.py
tallyhawley: you're right -- just pushed something to ensure that we only catch TypeError

jgranley (approved):

PR #593, opened 2023-12-19
[ENH][FIX][DOC] Add Dynaphos activation threshold

tallyhawley:
## Description

Adds a cortical tissue activation threshold to the Dynaphos model, to better match the paper. This is a settable parameter (a_thr) and limits when percepts are generated. This also prevents the currently observed behavior where, if a stimulus goes from high amplitude to 0, the generated phosphenes never fully fade.

Updates Dynaphos documentation to match, and adds an example allowing us to reproduce a figure from the paper (with limitations).

## Type of Change

- [x] Documentation change

## Checklist


jgranley:
Looks great!

PR #601, opened 2024-01-25
[ENH] Add methods to initialize an EnsembleImplant from a list of coordinates

tallyhawley:
## Description

Adds two methods to initialize an `EnsembleImplant` from a set of coordinates. This should generally match the format of `Neuralink.from_neuropythy()`.

`EnsembleImplant.from_coords()` allows the user to pass an implant type (must subclass `ProsthesisSystem`) & a set of locations on the cortex/retina (in µm), or an xrange/yrange/xystep in µm. We then automatically create a set of implants of the given type at the desired locations.

`EnsembleImplant.from_cortical_map()` takes the same parameters as `from_coords()` in addition to a `vfmap` (must be a `CorticalMap` instance and implement `from_dva()`) and `region` (defaults to v1). Locations/coordinates are passed to this function in dva, and internally converted to µm (in the given region).

Possible expansion: `from_retinal_map()`. Not implemented currently due to lack of use-cases as retinal implants are generally stand-alone. Would take an `eye` parameter instead of `region`?

## Type of Change

## Checklist

### DIFF
--- a/pulse2percept/implants/ensemble.py
+++ b/pulse2percept/implants/ensemble.py
@@ -9,6 +9,49 @@ class EnsembleImplant(ProsthesisSystem):
     # Frozen class: User cannot add more class attributes
     __slots__ = ('_implants', '_earray', '_stim', 'safe_mode', 'preprocess')
 
+    @classmethod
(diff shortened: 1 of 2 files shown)


jgranley:
Looks great!

PR #616, opened 2024-05-05
[ENH] Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within setup.cfg
+---
+name: Codespell
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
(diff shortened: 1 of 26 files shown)


jgranley:
This looks really cool, thanks for adding. Need to verify with the team but I would be in support of adding.

yarikoptic:
Yes, you can run with -w to perform non ambiguous fixes like I did in https://github.com/pulse2percept/pulse2percept/pull/616/commits/d5e3114fb16c0d0b2ff3f92f63f958181070cce2 . For ambiguous would need interactive mode like I did in subsequent commit I believe with -i 3.
Workflow does only detection ATM. 

You can add more skips to config - I guess you have some files which aren't in git and thus get more hits.

mbeyeler:
Thanks for proposing this and for fixing the spelling errors you found.

We are cleaning up older PRs and have decided not to add a separate codespell workflow at this time. We would prefer to keep the CI setup relatively lean and avoid adding another dependency and required check unless it addresses a recurring problem for the project.

This PR also predates substantial changes to our packaging and CI configuration, so it would need to be reworked rather than merged in its current form. I’m going to close it rather than leave it open indefinitely.

yarikoptic:
your call... note that I think I did address all questions etc so AFAIK it was ready back then . let me know if I should reincarnate it up to date

PR #619, opened 2024-05-18
[MNT][FIX] Update GitHub actions for MacOS

tallyhawley:
## Description

Updates the GitHub actions workflows so that Python is actually installed and setup properly on MacOS runners.

Excludes 3.8 and 3.9 for latest mac since there is no runner, and adds in macos-13 to test these versions. Also installs ffmpeg for latest mac.

## Type of Change


jgranley (approved):
Looks great, I think the test precision changes are fine, that still shouldn't be noticeable that small

PR #626, opened 2024-06-11
[ENH] Add temporal torchmodel

tallyhawley:
## Description

Adds torch backend for FadingTemporal, `TorchFadingTemporal`.

Refactors the TorchBaseModel `forward` signature to add a `state` kwarg for temporal models & to make the `e_locs` argument optional in the parent. Subclasses can choose which kwargs to use in the `forward`.

## Type of Change

## Checklist

### DIFF
--- a/pulse2percept/models/temporal.py
+++ b/pulse2percept/models/temporal.py
@@ -10,6 +10,7 @@ class TorchFadingTemporal(TorchBaseModel):
         super().__init__(p2pmodel)
         self.dt = torch.tensor(p2pmodel.dt, device=self.device)
         self.tau = torch.tensor(p2pmodel.tau, device=self.device)
+        self.thresh_percept = torch.tensor(p2pmodel.thresh_percept, device=self.device)
 
     def forward(self, stim, state=None, model_params=None):
         """_summary_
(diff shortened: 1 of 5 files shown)


pulse2percept/models/temporal.py
```diff
@@ -1,7 +1,82 @@
...
+            One frame of the predicted response
+        state : torch.Tensor
+            The internal state of the model after this pass. 
+            state[0] corresponds to ``bright``
+
+        """
+
+        tau = self.tau if model_params is None else torch.Tensor(model_params[0], device=self.device)
+
+        if state is None:
```
jgranley, line 42: Im not sure, but this might get compiled out when using torch.compile or onnx? 
Most if statements usually are. If this is the case might have to require that this is passed in 0's on the first iteration and not allow it to be none.

tallyhawley: that's definitely an option. i forgot that torch.compile/torch fx has issues with data-dependent control flow. i'll update it ASAP

jgranley:
I'm a little confused why this is failing right now. Theres an easy fix to the numpy array not being c-contiguous (just copy it with order='c' in predict_percept), but the test where this is failing shouldnt even be running. We pass --benchmark-disable to pytest (pyproject.toml) , and it is skipping most of the other benchmarks, so i'm not sure why its still running this one. Eventually I'll change the workflow to also include benchmarking, but maybe for now we can look into why this test is even being ran?

mbeyeler:
A lot has changed, and even more is up in the air about how to move forward, but I will merge this for now

PR #635, opened 2024-06-20
[ENH][MNT] Update for Numpy 2.0

jgranley:
Numpy 2.0 breaks p2p in various places, mostly because operations involving python scalers now get type auto promoted to float64, where numpy is expecting to always use float 32 (and compiled cython expects float 32)

Note that in order to run with numpy 2.0, ALL packages must be updated, since compiled 1.0 headers no longer work with 2.0. However, p2p should still work with either 2.0 or 1.0, as long as every installed package has the correct version.


jgranley:
This is ready for numpy 2 support, but tests wont pass until neuropythy/pimms is made [numpy 2 compatible](https://github.com/noahbenson/pimms/pull/6)

jgranley:
The tests all pass locally, i think this is another weird thing with testing inside / outside of the install directory

mbeyeler:
Maybe we can target this for v0.10? One fell swoop with Py313 support right after it

mbeyeler:
About your error:

```
def test_fetch_url(tmp_data_dir):
        url1 = 'https://www.nature.com/articles/s41598-019-45416-4.pdf'
        file_path1 = os.path.join(tmp_data_dir, 'paper1.pdf')
        paper_checksum1 = 'e8a2db25916cdd15a4b7be75081ef3e57328fa5f335fb4664d1fb7090dcd6842'
>       fetch_url(url1, file_path1, remote_checksum=paper_checksum1)
```

I think Nature is blocking repeated requests to combat scrapers. All is needed here is a permanent URL with some kinda file download so we can compare against the checksum

jgranley:
@mbeyeler We can wait to merge for py313 to drop py39, but it also can be merged now without issue. Merging will just mean two additional lines for 3.9 in `pyproject.toml` will have to be changed with py313

Issue #571, opened 2023-10-26
[REF] Use pyproject.toml


jgranley:
Having the package use a pyproject.toml to set up the build, installation, and pip package would be ideal, would modernize the package, and would provide numerous other benefits such as simplified installation, cross platform compatibility, allows pulse2percept to be included in other projects requirements.txt (currently doesn't work because numpy and cython must be installed prior to pulse2percept), and simplified github actions (e.g. wheel build checks).

https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/

mbeyeler:
Closed by #644

PR #637, opened 2024-06-22
[ENH] Add benchmarking

### LINKED ISSUE #627: [ENH] Add time benchmarking with pytest-benchmark
Use pytest-benchmark to benchmark computation time of different models with different inputs.

jgranley:
Adds timing benchmarking using pytest-benchmark for all models.

This is disabled by default, but can be enabled by passing `--benchmark-enable` to pytest.

Some options could be removed eventually. Right now every model is benchmarked with all engines (serial, cython, and torch), as well as torch on cpu and gpu, and torch with and without compilation. 

Closes #627 

This also adds a barebones pyproject.toml. Eventually we should migrate more setup, installation, and configuration to here (#571)

### DIFF
--- a/pulse2percept/models/beyeler2019.py
+++ b/pulse2percept/models/beyeler2019.py
@@ -296,8 +296,6 @@ class AxonMapSpatial(SpatialModel):
             'ignore_pickle': False,
             # Use the Watson transform for dva <=> ret:
             'vfmap': Watson2014Map(),
-            # whether or not to compile torch models
-            'compile': False
         }
         return {**base_params, **params}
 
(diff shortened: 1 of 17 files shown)


pulse2percept/models/beyeler2019.py
```diff
@@ -295,7 +295,9 @@ def get_default_params(self):
             # You can force a build by ignoring pickles:
             'ignore_pickle': False,
             # Use the Watson transform for dva <=> ret:
-            'vfmap': Watson2014Map()
+            'vfmap': Watson2014Map(),
+            # whether or not to compile torch models
+            'compile': False
```
jgranley, line 300: Move to base class

PR #656, opened 2025-02-13
[MNT] Improve macOS / Windows compatibility

mbeyeler:
Some long-overdue maintenance for our cross-platform compatibility:

- [x] rewrites `utils/_fast_array.pyx` to use strictly C-level code
- [x] makes sure compilers use C++ flags where appropriate
- [x] tweaks install procedure on `build.yml` and `wheels.yml` 
- [x] restricts `build.yml` to common platforms (e.g., "latest" OS, new & old Python)
- [x] extends `wheels.yml` to all supported platforms
- [x] improves installation instructions (pip vs source)
- [x] will warn if trying to install on an unsupported configuration (e.g., Python 3.13)
- [x] will warn if Visual Studio Build Tools are missing


mbeyeler:
This is madness. I think I'm gonna break it down into multiple smaller PRs

Issue #639, opened 2024-06-24
[ENH] Add stimulus encoders


jgranley:
Add naive / traditional stimulus encoders. These will be similar in design to image stimulus, but should 

- support multiple images (e.g. a call function which takes an image as input, and outputs a stimulus for the specified implant). 
- use `to_dva` instead of just resizing the image.  
- support single frames or videos. 
- support amplitude encoding, or modulating parameters of a biphasic pulse train
- be implemented in torch

PR #680, opened 2025-02-27
[ENH] Smarter reshape_stim for implants

mbeyeler:
Automatically converting `ImageStimulus` objects for irregularly spaced electrode grids is tricky. The current approach is to build a higher-resolution stimulus with step size equal to the smallest $\Delta x$ separation of electrode locations. However, this quickly runs out of memory if electrode offsets are small.

A better approach is to use interpolation, where the image is assumed to span the whole range of `[min_x, max_x, min_y, max_y]` electrode locations and the value at `x,y` is interpolated.

### DIFF
--- a/pulse2percept/implants/base.py
+++ b/pulse2percept/implants/base.py
@@ -2,16 +2,13 @@
    :py:class:`~pulse2percept.implants.RectangleImplant`"""
 import numpy as np
 from copy import deepcopy
-from functools import reduce
-from scipy.spatial import cKDTree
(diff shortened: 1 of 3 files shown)


jgranley:
Related is #639. I think the best way to do this is to use the visuotopy to sample the visual field location right at the electrode.

mbeyeler:
Right - this is actually tricky right now, because only the model knows about topography, not the implant (and certainly not the stimulus). But this would be possible with the informally discussed API changes

PR #686, opened 2025-03-02
[MNT] Remove support for joblib and dask

### LINKED ISSUE #664: [MNT] Drop support for dask, joblib?
These were hot in 2019, but now we have Cython+OpenMP and more importantly, torch

mbeyeler:
Remove support for JobLib and Dask (closes #664 ). Most models now use either Cython, Jax, or Torch.


arokem:
The end of an era! 😄 

By the way, if in the future you find yourself needing similar functionality, we've recently found that [Ray](https://www.ray.io/) provides excellent performance for this type of thing (see https://nrdg.github.io/2024-dipy-parallelization/).

mbeyeler:
Haha, I was wondering if you're still getting bombarded with our notifications! 😆

Thanks for the suggestion - Ray looks promising, and we'll definitely keep it in mind for the future!

PR #738, opened 2025-09-05
[MNT] Colab-ish CI checks for install

mbeyeler:
This PR adds lightweight GitHub Actions workflows that emulate a Colab environment to catch install issues early:

- PR smoke test: On every push/PR, installs pulse2percept from the exact commit/branch using `pip install git+…`, checks runtime dependencies, and runs a basic import/`ScoreboardModel` instantiation.

- Daily drift check: Scheduled job that installs from both GitHub master and PyPI, in two profiles:
    - latest (no pre-pins) to detect upstream breakage.
    - colab-pins (NumPy/JAX/TensorFlow/PyTorch versions Colab ships) to mirror the real Colab runtime.

This ensures users’ common workflow (`!pip install git+…` or `!pip install pulse2percept`) continues to work smoothly in Colab.


mbeyeler:
Right idea, wrong method/approach

PR #868, opened 2026-08-28
[ENH] Add PRIMA optical stimulus encoding

mbeyeler:
Adds a device-specific encoding pipeline for `PRIMAPivotal`, mapping images and videos to 880 nm optical stimulation rather than electrical current.

- [x] adds optical power/irradiance units and a generic `Encoder` base class
- [x] adds `PRIMAEncoder` with binary and PWM grayscale encoding, projector timing, and documented device-envelope checks
keeps encoded optical waveforms lazy
- [x] allows `ScoreboardModel` to visualize normalized optical drive for PRIMA without pretending to model photovoltaic or retinal transduction
- [x] adds tests, docs, and a gallery example

Photovoltaic circuit and retinal-response modeling remain out of scope.

### DIFF
--- a/benchmarks/scenarios.py
+++ b/benchmarks/scenarios.py
@@ -15,14 +15,17 @@ first two correspond to these one-liners::
                             xrange=(-12, 12)).predict_percept(
         as_current(implant, p2p.stimuli.LogoBVL()))
 
-    implant = p2p.implants.PRIMAPivotal()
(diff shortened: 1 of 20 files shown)


mbeyeler:
The benchmark legit changed what it measured - now accounting for roughly 3x the work to accommodate the image -> optical path

Issue #852, opened 2026-08-24
Improve fit of Dynaphos model with published figure, overhaul docs


mbeyeler:
Using the protocol

```
BiphasicPulseTrain(
    freq=300,
    amp=100,
    phase_dur=0.17,
    stim_dur=166,
)
t_percept = np.arange(0, 700, 20)
```

we should be able to obtain their Figure 2. I think a model-specific method is justified here

```
model.predict_brightness(implant, t_percept=None)
```

so we can compare apples to apples.

A regression test should make sure our numbers are (and stay) accurate.

The gallery example should be reworked accordingly.

charliefay38221-sudo:
A partial answer to the regression-test half of this, from the other direction.

`dynaphos-experiments/brightness/results/brightness_estimate.csv` publishes the model's own predicted brightness alongside the tissue activation that produced it. Brightness is a pure function of activation in this model, so that pair can pin the calibrated sigmoid **without** having to reconstruct the stimulation protocol first:

```
 amp (uA)    activation     paper       p2p   abs diff
       10    0.0000e+00  0.116537  0.116537   1.78e-08
       30    2.5038e-08  0.175650  0.175650   1.03e-08
       50    1.0713e-07  0.506540  0.506540   3.86e-08
       80    2.3027e-07  0.915642  0.915643   6.70e-08
      100    3.1236e-07  0.981235  0.981235   4.61e-08
```

Worst absolute difference across all ten published rows is 6.7e-08, i.e. float32 noise. So the brightness stage here already reproduces the published model exactly, and that is worth locking down independently of the Figure 2 work.

```python
# Digitised from van der Grinten et al. (2024), via
# dynaphos-experiments/brightness/results/brightness_estimate.csv
PUBLISHED_BRIGHTNESS = [
    (0.0,                    0.11653672158718109),
    (2.5037856232756894e-08, 0.17564991116523743),
    (6.608350844317101e-08,  0.31865283846855164),
    (1.0712917486443985e-07, 0.5065404176712036),
    (1.481748199694266e-07,  0.6925985217094421),
    (1.8922051481240487e-07, 0.8317987322807312),
    (2.302661812336737e-07,  0.9156424403190613),
    (2.713118192332331e-07,  0.9597164392471313),
    (3.123574572327925e-07,  0.9812352061271667),
]

def test_brightness_matches_published_estimates():
    model = DynaphosModel()
[...]

PR #873, opened 2026-08-30
Fix Dynaphos phosphene size: sigma is P/4, not P/2

charliefay38221-sudo:
`sigma = P / 2` becomes `sigma = P / 4`. The +/-2 sigma interval should span the phosphene size P, so rendered phosphenes were twice their intended standard deviation.

Rebased on current master and cut back to the minimum: one behavioural line, plus the pinned test values it invalidates. The grid steps in those tests are reduced because correctly sized phosphenes fall below one pixel at the previous resolution.

Full suite locally: 28 failures before and after, none introduced.

### DIFF
--- a/pulse2percept/models/cortex/dynaphos.py
+++ b/pulse2percept/models/cortex/dynaphos.py
@@ -439,7 +439,7 @@ class DynaphosModel(BaseModel):
             D = 2 * np.sqrt(amp / K) # mm
             P = (D / M) # dva
             # calculate sigma for gaussian (only update sigma if amplitude > 0)
-            sigma = np.where(amp > 0, np.clip(P / 2, 1e-22, None), sigma) 
+            sigma = np.where(amp > 0, np.clip(P / 4, 1e-22, None), sigma) 
             # get activation (Ieff converted from uA to A)
             A = A + ((-A / tau_act_s) + Ieff * _A_PER_UA) * dt_s
             # get brightness
--- a/pulse2percept/models/cortex/tests/test_dynaphos.py
+++ b/pulse2percept/models/cortex/tests/test_dynaphos.py
@@ -117,14 +117,14 @@ def test_temporal_predict():
 
     # Test that default models give expected values
     orion = DynaphosModel(implant=Orion(), implant_position=(15, 0) * mm,
-                          step=0.1, dt=20).build()
+                          step=0.01, dt=20).build()
     percept = orion.predict_percept(
         {'55': BiphasicPulseTrain(freq=300, amp=100, phase_dur=0.17)})
-    npt.assert_equal(np.sum(percept.data > 0.0122), 147)
-    npt.assert_equal(np.sum(percept.data > 0.0375), 96)
-    npt.assert_equal(np.sum(percept.data > 0.3305), 49)
-    npt.assert_equal(np.sum(percept.data > 0.8451), 39)
-    npt.assert_equal(np.sum(percept.data > 0.8883), 9)
+    npt.assert_equal(np.sum(percept.data > 0.0122), 4026)
+    npt.assert_equal(np.sum(percept.data > 0.0375), 2806)
+    npt.assert_equal(np.sum(percept.data > 0.3305), 907)
+    npt.assert_equal(np.sum(percept.data > 0.8451), 172)
+    npt.assert_equal(np.sum(percept.data > 0.8883), 90)
 
 def test_deepcopy_Dynaphos():
     original = DynaphosModel(implant=Cortivis())
@@ -391,7 +391,7 @@ def test_location_noise():
                                             phase_dur=0.17)}
     kwargs = dict(implant_position=(20, -5) * mm,
                   xrange=(-4, 4), yrange=(-4, 4),
-                  step=0.05)
+                  step=0.02)
     plain = DynaphosModel(implant=implant, **kwargs).build()
     expected = plain.predict_percept(source).data
 
@@ -424,7 +424,7 @@ def test_location_noise_crosses_meridian():
     # Choose an electrode/offset pair that crosses the vertical meridian.
     implant = Implant(ElectrodeArray([DiskElectrode(-25000, 2000, 0, 100)]))
     source = {0: BiphasicPulseTrain(freq=300, amp=200, phase_dur=0.17)}
-    kwargs = dict(xrange=(-4, 4), yrange=(-4, 4), step=0.05)
+    kwargs = dict(xrange=(-4, 4), yrange=(-4, 4), step=0.02)
     plain = DynaphosModel(implant=implant, **kwargs).build()
     canonical = plain.predict_percept(source)
     npt.assert_array_less(0, _brightest_dva(canonical, plain.grid)[0])
--- a/pulse2percept/models/tests/test_api_equivalence.py
+++ b/pulse2percept/models/tests/test_api_equivalence.py
@@ -66,10 +66,11 @@ REFERENCE = {
                  3.807037961360792, 0.5194367010755934, 1.0132546632537747,
                  (5.0853018679140565, 27.047305434768763,
                   6.941999430159145, 49.25247073610444)),
-    'dynaphos': ((7, 7, 6), (0.0, 100.0),
-                 3.864256768792984e-06, 1.0946714610327035e-06,
-                 3.800738015684018e-12,
-                 (2.0, 4.0, 1.0, 1.0)),
+    'dynaphos': ((121, 121, 6), (0.0, 100.0),
+                 7.886836596880443, 0.8272089958190918,
+                 3.4563948838340095,
+                 (40.9433498104171, 1676.7244805318699,
+                  26.316482385133106, 692.9332594630685)),
     'scene_gaze': ((13, 17, 1), None,
                    1663.1452019751928, 36.72337341308594, 43973.7099062507,
                    (6.000000013081055, 40.855445551864825,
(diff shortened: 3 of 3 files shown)


charliefay38221-sudo:
Pushed a second commit: `test_dynaphos_prediction_is_unchanged` in `test_api_equivalence.py` also pinned Dynaphos output, and I had missed it initially.

Two things there worth your attention, one of which is independent of this PR.

**The existing reference does not reproduce on my machine.** On stock `master`, arm64 with numpy 2.x, that test already fails: `data.sum()` evaluates to exactly `0.0` against the pinned `3.864256768792984e-06`, a relative difference of 1. The cause is the same sub-pixel sampling issue: at `step=1` a phosphene at that eccentricity is far smaller than one pixel, so the reference recorded the far tail of the Gaussian (peak `1.09e-06`, about a millionth of full brightness) rather than the prediction. Values that small are at the mercy of platform floating-point.

**With sigma corrected it gets worse before it gets better.** The same sample drops to `1.7e-24`, because halving sigma quadruples the exponent. Pinning that would be pinning noise.

So I moved that entry to `step=0.05`, where the values are order 1 (peak 0.827, total 7.89) and comfortably above the suite's `rtol=1e-5`. It costs a `(121, 121, 6)` array instead of `(7, 7, 6)`; the file still runs in about 3 s here.

Full suite locally: 24 failures both with and without this change, i.e. none introduced. All of them are missing-optional-dependency or network cases (`h5py`, video/ffmpeg, `test_fetch_url`). The only difference in either direction is the dynaphos entry above, which fails on `master` here and passes with the change.

I have not been able to check the new values against your CI environment, so that is worth a look before merging. If you would rather keep `step=1` and pin the tiny values, or drop that entry, say so and I will adjust.

charliefay38221-sudo:
Stronger evidence than the textual argument, from the authors' own published intermediate values.

`neuralcodinglab/dynaphos-experiments` ships `temporal_dynamics/results/predicted_temporal_dynamics.csv`, which contains a `predicted_sigma_` column:

```
predicted_sigma_ = 0.014079296630263844
predicted_brightness_ = 0.4636700295652317
```

That brightness value is the same `0.4636` already pinned in `bright_amp_ref` in `test_temporal_predict` here, so we are looking at the same model state.

The notebook that produced it (`temporal_dynamics/temporal_dynamics_Oct_22.ipynb`, cell 12) generates those columns by running the reference simulator directly:

```python
params = utils.load_params('../../config/params.yaml')
params['cortex_model']['dropout_rate'] = 0
params['default_stim']['pw_default'] = 100e-6
params['default_stim']['freq_default'] = 200
params['thresholding']['use_threshold'] = False
coordinates_visual_field = get_visual_field_coordinates_probabilistically(params, n_phosphenes=1)
simulator = PhospheneSimulator(params.copy(), coordinates_visual_field)
```

It overrides dropout, pulse width, frequency, frame rate and thresholding, but **not** `radius_to_sigma`, which stays at its config default of `0.5`. That is the P/4 path: dynaphos computes `sigma = sqrt(I/K) * radius_to_sigma / M`, and since `sqrt(I/K)` is the radius and `P = 2*sqrt(I/K)/M`, that is `P/4`.

So the sigma published alongside the paper is the P/4 convention, independently of how anyone reads the sentence in the methods.

As a sanity check on plausibility, inverting that published sigma for the magnification each convention would require:

| amp (uA) | M for P/4 | ecc | M for P/2 | ecc |
|---|---|---|---|---|
| 30 | 7.49 | 1.56 deg | 14.97 | 0.41 deg |
| 45 | 9.17 | 1.14 deg | 18.34 | 0.19 deg |
| 60 | 10.59 | 0.88 deg | 21.18 | 0.07 deg |

P/4 puts the phosphene around 1 degree of eccentricity, which is where you would expect an implanted electrode. P/2 requires it to sit essentially at the foveal centre.

I should say plainly that I have not identified which electrode and amplitude that particular row corresponds to, so the table is a plausibility check rather than a derivation. The notebook evidence above is the part I would rely on.

charliefay38221-sudo:
Closing the gap I left in the previous comment: I have now identified the amplitude, so this is no longer a plausibility argument.

The notebook that generated `predicted_temporal_dynamics.csv` sets `STIM_AMPLITUDE = 90e-6` for that experiment, with `pw_default = 100e-6`, `freq_default = 200`, `fps = 256`, `dropout_rate = 0` and `use_threshold = False`. Sigma in this model depends only on amplitude, and the published `predicted_sigma_` is constant across the file, consistent with the constant 90 uA during stimulation.

With the amplitude known, each convention implies a required cortical magnification, and only one of them is achievable:

| convention | required M (mm/dva) | required eccentricity |
|---|---|---|
| `sigma = P/4` | 12.968 | 0.569 deg |
| `sigma = P/2` | 25.935 | **no solution exists** |

The maximum magnification anywhere in the visual field, at the foveal centre, is **22.922 mm/dva** under this model's own `cortex_model` parameters. `P/2` requires 25.935, which is larger than the model can produce at any eccentricity. So there is no electrode position, anywhere, at which the `P/2` convention reproduces the sigma published with the paper.

Confirmed by direct simulation rather than algebra. Running the reference simulator with the notebook's exp-2 settings, one phosphene at 0.569 deg, 90 uA:

```
dynaphos sigma      0.0140739288 dva
paper published     0.0140792966 dva
relative difference 3.81e-04
```

The residual is my rounding of the eccentricity to three decimals.

So the case now rests on three independent legs, and the weakest one is no longer load-bearing:

1. the two implementations differ by exactly 2.00x, measured
2. the methods sentence requires the +/-2 sigma interval to equal P, hence sigma = P/4
3. the sigma published alongside the paper is reproduced by the P/4 path, and is unreachable under P/2 at any eccentricity

If you disagree with (2), which was always the most arguable, (1) and (3) stand on their own.

charliefay38221-sudo:
One more piece, this time checking whether the size equation was the *only* divergence rather than just *a* divergence.

I ran a differential test of the brightness cascade between this model and the reference implementation, driving both with matched stimulation and comparing the brightness time series. Brightness and activation are position-independent in both, so no geometry matching is needed.

With this PR applied:

| condition | worst relative difference |
|---|---|
| 30 uA / 300 Hz / 170 us (sub-threshold) | 0.00% |
| 80 uA / 300 Hz / 170 us | 1.05% |
| 200 uA / 300 Hz / 170 us | 0.42% |
| 80 uA / 200 Hz / 100 us | 0.00% |

So after the sigma correction the two implementations agree to about 1%, and the size equation appears to have been the only material divergence. The residual is consistent with integration details (the reference sub-steps its activation update when the decay rate exceeds the frame rate; this model takes a single Euler step per `dt`).

Two things I had to reconcile to get there, both of which look relevant to #852.

**There is a one-frame offset between the two time bases.** `t_percept=0` here corresponds to the state *before* the first activation update, whereas the reference's loop updates and then reads. Every condition above needed a shift of one frame to line up, except the sub-threshold ones where both are zero throughout. Anyone comparing a time series against Figure 2 will need to account for that, and it is easy to mistake for a model error.

**Thresholding differs structurally, not just numerically.** `a_thr` here is a scalar. The reference draws a per-electrode threshold from a distribution, which is what makes its `use_threshold` behaviour probabilistic across electrodes. For the comparison above I pinned the reference to a scalar threshold to match. Not a defect in either, but the two will not agree electrode-by-electrode on a populated array unless that is accounted for.

Caveats: four conditions, one electrode, and the alignment shift was fitted rather than derived. I would treat the 1% as "no remaining divergence of the size we just fixed" rather than as a precise agreement figure.

charliefay38221-sudo:
Checked what this does to the gallery example (`examples/models/plot_dynaphos.py`), since it renders Dynaphos output in the docs and I did not want that to change silently.

**The Fig. 3 brightness-over-time block is essentially unaffected:**

| amp (uA) | before | after |
|---|---|---|
| 50 | 0.5397 | 0.5374 |
| 70 | 0.8619 | 0.8593 |
| 100 | 0.9866 | 0.9845 |

About 0.2%. That block stimulates every electrode of the implant and takes the max over the whole raster, so the reading is dominated by a peripheral phosphene that the grid samples adequately. The sampling penalty I described earlier does not reach it.

That is also the explanation for something that otherwise looks inconsistent: rasterised brightness loses 2-14% for a **single foveal** electrode but almost nothing for a **full array**, because with an array something is always well sampled. Worth knowing before trusting any single measurement of this.

**The spatial figure does change visibly:**

| | before | after |
|---|---|---|
| peak brightness | 0.9907 | 0.9886 |
| lit pixels (>0.05) | 53 | 16 |

On the default 41x41 grid that is roughly a third of the previously lit area, so the rendered percept will look noticeably sparser. Peak brightness is unaffected; it is purely that the phosphenes are now their correct smaller size and the default grid resolves them less well.

I have not touched the example in this PR, since you mentioned in #852 that it wants reworking anyway and I would rather not collide with that. But if you want the docs image to look comparable, that example likely wants a finer `step` rather than any change to the model.

mbeyeler:
Charlie and Claude, I appreciate you taking this on. The problem was identified correctly and the core fix looks right. My concern is that the implementation has grown quite a bit around what is essentially a one-line correction.

For example, `test_phosphene_size_matches_the_analytical_sigma` is a good idea, but it builds the model with `step=.005`, the default ranges (a 2001x2001 grid), and a 1-second pulse train (which Dynaphos expands to a 780 MB array). The documentation also overstates the "95%" argument. `test_dynaphos_prediction_is_unchanged()` no longer establishes API equivalence at all, so I think it should probably just be removed.

Our contribution guidelines allow AI-assisted work, but they still require a human contributor to understand, review, and take responsibility for the submitted code. At this point I can't tell which of these design choices you personally consider necessary versus what your AI assistant generated along the way.

Could you please go through the PR yourself and reduce it to the smallest change you believe is needed? I would expect roughly:
- the `P / 4` fix
- one small analytical regression test
- concise documentation
- only the minimal handling needed for existing tests that were invalidated by the now-correct behavior

Once that is done, I'm happy to review the resulting code in detail.

Many thanks!

mbeyeler:
Howdy! Do you still plan on de-Claude-ifying this? Or do you want me to take over?

charliefay38221-sudo:
Rebased on master and stripped back to the minimum: 15 lines across 3 files, no added comments, no new tests. Sorry for the noise on the first pass.

mbeyeler:
Thanks, Charlie. As we discussed, I'm happy to take over. Since I can't push to your branch, I'll open a fresh branch/PR on our side.

Really appreciate you finding and digging into this bug in the first place!

